PR #1, opened 2020-09-12
Adding tests and library to use connectors

ilblackdragon:
Testing NEAR <> Ethereum token connector, while using Mock prover on both sides.

This starts to use rainbow-bridge-lib for various tools and components.


Kouprin (commented):
this commit contains `package-lock.json`, it's npm-related thing
can we sync in what is our standard - yarn or npm?

ilblackdragon:
I have no idea, but because my internet is somewhat not stable, sometimes one of them doesn't work.

Kouprin:
> yarn install v1.21.0

yarn install v1.22.4 finishes well for me
could you please update to 1.22.4?

ilblackdragon:
It's not the problem with version. Your internet is very different from my internet. 
I'll remove the package-lock.json for now.

index.js
```diff
@@ -0,0 +1,179 @@
...
+        changeMethods: [
+            { methodName: 'deposit', inputFieldType: 'Proof', outputFieldType: null },
+            // {'deploy_bridge_token'}, 
+            // {'lock'}, 
+            // {'unlock'}],
+        ]
+    });
+}
+
+async function createEthConnector(web3, config) {
```
chadoh, line 67: I'd rather see destructuring used here instead of passing an opaque `config` so that it's more obvious what values need to be passed in

TypeScript would be even better! 😄

index.js
```diff
@@ -0,0 +1,179 @@
...
+            ['log_entry_data', ['u8']],
+            ['receipt_index', 'u64'],
+            ['receipt_data', ['u8']],
+            ['header_data', ['u8']],
+            ['proof', [['u8']]],
+        ],
+    },
+}
+
+async function createNearConnector(masterAccount, config) {
```
chadoh, line 42: It seems like there are two possible use-cases for this function, and I want to see if I understand them correctly:

1. from a server/command line environment, where you have a proper "master account" that can create a new sub-account and deploy a connector contract
2. a frontend app that will pass in the signed-in user's account and a `nearConnectorId` for an existing connector

Please correct this understanding if it's off

I'm trying to figure out what values to pass to this from my frontend app at https://github.com/near-examples/erc20-to-nep21/pull/1/files

mfornet:
There is an ongoing process to create a library to use connectors on the frontend-repository https://github.com/near/rainbow-bridge-frontend/pull/75.

cc @paouvrard

PR #3, opened 2020-09-21
Make ERC20Locker ERC223 compatible

robert-zaremba:
There are few ERC223 tokens out there and we would like to support them on the NEAR side. 
However, ERC223 has this rigorous check:  it doesn't allow token transfers to contracts that don't support token receiving and handling.
This PR implements a dummy `tokenFallback` function to satisfy ERC223 requirements.


robert-zaremba:
Copying changes form https://github.com/near/rainbow-bridge-sol/pull/21

mfornet (approved):
LGTM

PR #7, opened 2020-10-20
add get_bridge_token_account_id info to the README

### LINKED ISSUE #6: Implement a method "is token deployed"
According to @chadoh it's very good to have a method that says that is the token already deployed. Otherwise we suggest a weird error-first experience to try to call a contract method to know is it deployed.

Kouprin:
Fixes #6.

It seems that `get_bridge_token_account_id` already gives information about is token deployed or not. The only necessary step is to update README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,11 @@ impl BridgeTokenFactory {
     /// Expects ~35N attached to cover storage for BridgeToken.
     #[payable]
     pub fn deploy_bridge_token(address: String);
+
+    /// Checks if Bridge Token has been successfully deployed with `deploy_bridge_token`.
+    /// On success, returns the name of NEP21 contract associated with given address (<hex(evm_address)>.<current_id>).
+    /// Otherwise, returns "token do not exists" error.
+    pub fn get_bridge_token_account_id(&self, address: String) -> AccountId;
 }
 
 struct BridgeToken {


README.md
```diff
@@ -59,6 +59,11 @@ impl BridgeTokenFactory {
     /// Expects ~35N attached to cover storage for BridgeToken.
     #[payable]
     pub fn deploy_bridge_token(address: String);
+
+    /// Checks if Bridge Token has been successfully deployed with `deploy_bridge_token`.
+    /// On success, returns the name of NEP21 contract associated with given address (<hex(evm_address)>.<current_id>).
+    /// Otherwise, returns "token do not exists" error.
+    pub fn get_bridge_token_account_id(&self, address: String) -> AccountId;
```
chadoh, line 66: Does this return Borsh- or JSON-encoded data?

chadoh (approved):
Great, thanks! See one quick question below. It might be worth updating the README with this extra info before merging.

README.md
Kouprin: It seems all borsh-serialized methods has `#[result_serializer(borsh)]`. This method doesn't.
If we want to add info about borsh serialization, we should do it with all methods, (`finish_withdraw` for example). However, we want to keep README as light as possible to onboard developers quickly. Personally, I don't think we need to differentiate methods result serialization in README.

PR #10, opened 2020-10-26
fix(contract): Call record after proof verification

### LINKED ISSUE #5: Token Factory can be abused
I see an issues with the function `deposit` from `BridgeTokenFactory`.

This function receives as a parameter a proof that contains relevant information of an event that took place on Ethereum side, and relevant tokens are locked, so they can be safely minted in the NEAR side. The strategy to avoid double mints are storing all proofs in the state, and checking no repeated proof is stored.

1. This approach has a clear drawback, which is that state might blow up, just from usage by legitimate users.
2. Inside the `deposit` function, the first step is calling [`record_proof`](https://github.com/near/rainbow-token-connector/blob/master/bridge-token-factory/src/lib.rs#L130). This means, that any proof that is submitted, disregarding if it is correct or not, will be stored inside the contract. I think we should only store valid proofs, with valid merkle-roots from the Ethereum blockchain.
3. Malicious user can prevent another user to mint its tokens if it call `deposit` with an invalid proof. In this case the deposit call will fail, but before failing, will store proof information with `record_proof` inside the contract, so this proof can't be used anymore.

mfornet:
Fixes #5

Test plan
=======

Simulation test that showcase previous attack, and should not be possible after fixing.


bridge-token-factory/src/lib.rs
```diff
@@ -263,12 +317,14 @@ impl BridgeTokenFactory {
             &self.prover_account,
             NO_DEPOSIT,
             env::prepaid_gas() / 4,
-        ).then(ext_self::finish_unlock(
+        )
+        .then(ext_self::finish_unlock(
             event.token,
             event.recipient,
             event.amount,
+            proof_1,
```
Kouprin, line 325: Is it possible to do something like
```
ext_prover::verify_log_entry(
            proof.log_index,
            proof.log_entry_data,
            &self.prover_account,
            NO_DEPOSIT,
            env::prepaid_gas() / 4,
        ).then(record_proof(proof)).then(finish_unlock)
```
?

bridge-token-factory/tests/token_transfer.rs
```diff
@@ -237,6 +243,53 @@ fn test_eth_token_transfer() {
     assert_eq!(token.get_total_supply(&mut runtime), "900");
 }
 
+#[test]
+fn test_with_invalid_proof() {
```
Kouprin, line 247: Should this test panic?

mfornet: I already fix the test, it should not panic.

bridge-token-factory/src/lib.rs
mfornet: @Kouprin we can do it, and probably a better way to do it is with:

```rust
ext_prover::verify_log_entry().then(finish_unlock).and(record_proof)
```

so both calls are resolved in parallel. But it is kind of weird to do it, since both functions belong to the same contract, and we must split gas between both, while we can just do it everything inside a single call.

Notice that the call to `record_proof` inside `finish_unlock` is not a cross-contract call, but just an inner call to one of the functions.

Kouprin (approved):
the logic LGTM

PR #11, opened 2020-10-27
refactor: improve deploy_bridge_token error; lint

chadoh:
Most of the noise here comes from using [rustfmt] like a good Rustacean

The most substantive change comes from specifying the needed deposit in the `deploy_bridge_token` error message

Finally, I also added some human-readable comments next to some large numbers

  [rustfmt]: https://doc.rust-lang.org/book/appendix-04-useful-development-tools.html


frol (approved):

Kouprin:
@chadoh could you also run a linter at all rs-files in the repo? It would be more convenient to have all files fixed.

chadoh:
@Kouprin done in #12

PR #12, opened 2020-10-27
refactor: run 'cargo fmt' in all Rust directories

chadoh:
[rustfmt] can be added to most editors, and can also be run using `cargo fmt` within a specific Rust project

  [rustfmt]: https://doc.rust-lang.org/book/appendix-04-useful-development-tools.html


frol (approved):

Kouprin:
Thank you! We need to add pre-commit hooks on it.

chadoh:
Precommit works. Maybe a CI step is better

PR #14, opened 2020-11-13
fix(contracts): Remove ambiguity in results value

mfornet:


erc20-connector/contracts/BridgeTokenFactory.sol
```diff
@@ -79,6 +79,8 @@ contract BridgeTokenFactory is ERC20Locker {
 
     function _decodeLockResult(bytes memory data) internal pure returns(LockResult memory result) {
         Borsh.Data memory borshData = Borsh.from(data);
+        uint8 memory flag = borshData.decodeU8();
```
abacabadabacaba, line 82: Remove `memory`, `uint8` is a value type so it doesn't require storage specifier.

erc20-connector/contracts/ERC20Locker.sol
```diff
@@ -49,6 +49,8 @@ contract ERC20Locker is Locker {
 
     function _decodeBurnResult(bytes memory data) internal pure returns(BurnResult memory result) {
         Borsh.Data memory borshData = Borsh.from(data);
+        uint8 memory flag = borshData.decodeU8();
```
abacabadabacaba, line 52: Same here.

erc20-connector/contracts/BridgeTokenFactory.sol
mfornet: 🆗 

I still don't understand how is memory managed in solidity, I used `memory` from my understanding after reading here:
https://solidity.readthedocs.io/en/develop/introduction-to-smart-contracts.html#storage-memory-and-the-stack

So from you comment I understand that it is similar to rust traits `Copy` and `Clone`, in this case some basic types (such as `uint8`) is `Copy` so its value is not stored in `storage` but instead is directly stored in `memory`, without needing to specify it.

abacabadabacaba (approved):

bridge-token-factory/src/lib.rs
```diff
@@ -33,6 +33,12 @@ const TRANSFER_FROM_GAS: Gas = 10_000_000_000_000;
 
 const TRANSFER_GAS: Gas = 10_000_000_000_000;
 
+#[derive(Debug, Eq, PartialEq, BorshSerialize, BorshDeserialize)]
+pub enum ResultType {
```
nearmax, line 37: I would rather make it
```rust
pub enum ResultType {
  Withdraw(u128, [u8; 20], [u8; 20]),
  Lock(String, u128, [u8; 20]),
}
```
but I guess it is a matter of taste :)

erc20-connector/dist/BridgeTokenFactory.full.sol
```diff
@@ -529,7 +529,7 @@ library Borsh {
         }
     }
 
-    function decodeBytes20(Data memory data) internal pure returns(bytes20 value) {
+    function decodeBytes20(Data memory data) internal pure shift(data, 20) returns(bytes20 value) {
```
nearmax, line 532: I wonder where did `shift(data, 20)` come from? This looks suspicious. @abacabadabacaba WDYT?

nearmax (approved):

PR #16, opened 2020-11-25
fix(rust-contract): Bridge-Token functions are now payable

mfornet:
Add more robust script to compile contracts deterministically

Test plan
=======

- Deploy contracts locally, transfer tokens from Ethereum to NEAR, transfer tokens between different accounts in NEAR.


frol (approved):

bridge-token-factory/build.sh
```diff
@@ -1,5 +1,17 @@
-#!/bin/bash
+#!/usr/bin/env bash
+
+# Exit script as soon as a command fails.
+set -e
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
abacabadabacaba, line 6: Why not just `DIR="${0%/*}"`?

abacabadabacaba (approved):

bridge-token-factory/build.sh
frol: I like @mfornet version better since it uses a tool (`dirname`) that handles path properly while bash incantation may break things out unexpectantly.

mfornet: That line is actually not mine, iirc it is from @ailisp, and is implemented in this way to work well both locally and in CI

abacabadabacaba: @frol There are several problems with this approach. First, it creates two new processes just to determine the directory. Second, if `cd` command fails, the script doesn't exit. Third, if the name of the directory ends with a newline, it will be stripped. Also, if it starts with a dash, both `dirname` and `cd` will interpret it as an option, which will also break it.

It is a general problem that shell scripts tend to break when used with unexpected paths or when they encounter unexpected errors. A more robust solution would be:
```sh
if [[ "$0" == "${0%/*}" ]]; then
    DIR=.
else
    DIR="${0%/*}"
fi
```
This should work for all paths, even ones that contain unexpected characters.

PR #17, opened 2020-11-30
feat: add `withdraw` to Factory

chadoh:
To minimize the number of contracts for which a browser user will need a FunctionCall access key, this adds a `withdraw` function to BridgeTokenFactory and an associated `withdraw_internal` to BridgeToken.  `withdraw_internal` follows the pattern of similar BridgeToken functions, and is only callable by the BridgeTokenFactory that deployed this BridgeToken.


chadoh:
I would love to add tests for this new functionality, but I can't find many existing tests on which to base them and I'm not sure the best way to add them.

abacabadabacaba:
I think that there is a security assumption that to be able to transfer tokens from their account, a user needs access to the token contract. AFAIU, you propose to allow users to transfer any bridged tokens by calling a function on the factory contract instead, which breaks this assumption, and I don't like that.

chadoh:
In [BridgeToken::withdraw](https://github.com/near/rainbow-token-connector/blob/16937201f21c19251d1fd91a19d4341519eccc90/bridge-token/src/lib.rs#L76-L86):

```rs
    pub fn withdraw(&mut self, amount: U128, recipient: String) -> Promise {
        self.token.burn(env::predecessor_account_id(), amount.into());
        ext_bridge_token_factory::finish_withdraw(
            amount.into(),
            recipient,
            &self.controller,
            NO_DEPOSIT,
            env::prepaid_gas() / 2,
        )
    }
```

New [BridgeTokenFactory::withdraw](https://github.com/near/rainbow-token-connector/blob/16937201f21c19251d1fd91a19d4341519eccc90/bridge-token-factory/src/lib.rs#L207-L225):

```rs
    pub fn withdraw(&mut self, erc20_address: String, recipient: String, amount: U128) -> Promise {
        let _ = validate_eth_address(erc20_address.clone());
        let _ = validate_eth_address(recipient.clone());
        assert!(self.tokens.contains(&erc20_address), "Such BridgeToken does not exist.");

        let sender = env::predecessor_account_id();

        ext_bridge_token::withdraw_internal(...)
    }

```

Unless I'm missing something, both of these interfaces are callable by anyone, with equivalent results.

abacabadabacaba:
I mean that if there is a system which limits which contracts a user can call (which can be implemented using limited access keys, or a contract that will perform specific calls on the user's behalf, or by other methods), that system can be assuming that sending a token from the user's account is not possible without making calls to the token contract. This assumption is pretty reasonable, after all. For example, a UI which lets a user authorize a transaction can identify transactions that call a token contract and alert the user that the transaction accesses their tokens. If it is possible to send a token without making a call to the token contract, such checks can be bypassed.

mfornet:
Closing this since we are keeping our current interface.

Issue #28, opened 2021-02-16
Check contracts are compiled deterministically


mfornet:
Checks that contracts are compiled deterministically similar to how it was done in: https://github.com/near/rainbow-bridge/issues/459

PR #37, opened 2021-03-02
fix: Withdraw function is now payable

mfornet:
This is necessary so that users need to sign all
withdraw transactions.

This change requires an update on the frontend side. 

Fixes: https://github.com/near/rainbow-bridge-frontend/issues/106
Fixes: #28 

Test Plan
=========
CI pass


bridge-token/src/lib.rs
```diff
@@ -52,6 +52,7 @@ impl BridgeToken {
         self.token.mint(account_id, amount.into());
     }
 
+    #[payable]
```
djsatok, line 55: Should we return back the deposit? We don't need it to keep it

PR #38, opened 2021-03-02
Check that number of locked tokens isn't > 2^128

### LINKED ISSUE #34: Add checks on the amount of bridged tokens: they need to be lower than u128_max

mfornet:
Fixes #34

Testplan
========
- TockenLocker.js @ should not exceed max token limit


djsatok:
@mfornet 
Contract update LGTM, need to fix tests

djsatok (approved):
LGTM, please merge

PR #42, opened 2021-03-10
Add AdminControlled trait to BridgeToken

### LINKED ISSUE #41: Implement pausing of the `withdraw` for the bridge token
In order to correctly pause the transfers back we need to have an ability to pause `withdraw` method in all deployed bridge tokens. Potentially, this contract can be `AdminControlled`.

mfornet:
Allows pausing the withdraw function in the BridgeToken.

Closes #41


sept-en (approved):
LGTM. Shouldn't we merge this into `master` instead?

PR #51, opened 2021-04-01
fix: return `mint` promise from `deposit`

chadoh:
This transaction, https://explorer.near.org/transactions/erDQKaeYvmjKaPQz7deNKqLxZy8RFB3tFeHNut3yTRa, is marked as Successful. But you can see that the `mint` call failed.

```
The attached deposit is less than the mimimum storage balance
```

The change in the first commit here will correctly mark the whole `deposit` as failing, in such cases.

However, this contract still has a bug. Given that the `mint` step failed, the factory contract should roll back the `record_proof` step, so that I can try the full `deposit` again. Right now  when I retry the transaction, it fails a different way, https://explorer.near.org/transactions/FVtfxcz3SFZ5hhtbgnassHK454aAuBKfQ8yXofwRRAiT:

```
panicked at 'Event cannot be reused for depositing.', src/lib.rs:290:9
```

Also, I lost 100 DAI in the process of finding this bug. How do I get it back?


mfornet (approved):
This won't fix the issue, but at least it will show more meaningful error in the explorer, since the tx will be displayed as failed.

chadoh:
@mfornet when I build locally I do not see changes to any `wasm` files, but buildkite says we failed to check in these altered wasm files (why are build artifacts part of the git repository?).

Do you see these files when you build? Can you push the wasm files to fix the build?

mfornet:
> @mfornet when I build locally I do not see changes to any `wasm` files, but buildkite says we failed to check in these altered wasm files (why are build artifacts part of the git repository?).
> 
> Do you see these files when you build? Can you push the wasm files to fix the build?

We are adding this artifacts to the repository to guarantee deterministic compilation, and having access to actually deployed code. It is in our plan to update this workflow, and generate artifacts automatically in CI, but unfortunately it is not done yet. 

I'm re-compiling contracts now again. The issue was with a dependency that got outdated (not related to this PR). Not sure why previous commit failed, I will see if the output changes after this recompilation.

PR #53, opened 2021-04-20
Upgrade connector to send token to contract

mfornet:
- [ ] TODO: Refund tokens, and remove proof if the receiver didn't accept the EVM didn't accept the tokens. This may happen if the target ERC20 tokens is not deployed yet.
- [x] TODO: Measure how much gas is used by function, and use more accurate upper bounds.


bridge-token-factory/src/lib.rs
```diff
@@ -98,12 +102,67 @@ pub trait FungibleToken {
...
+            iter.collect::<Vec<&str>>().join(":"),
+            env::predecessor_account_id()
+        );
+        Recipient::ContractId {
+            evm_address,
+            message,
+        }
+    } else {
+        Recipient::AccountId(recipient)
+    }
```
abacabadabacaba, line 159:
```suggestion
    if let Some(index) = recipient.find(':') {
        Recipient {
            recipient[.. index].into(),
            Some(format!("{}:{}", recipient[index + 1 ..], env::predecessor_account_id())),
        }
    } else {
        Recipient { recipient, None }
    }
```
Also, what's the point of passing `predecessor_account_id`?
1. Because this function is called from `finish_deposit`, `predecessor_account_id` will always be the token factory account ID.
2. Even if this were called from within `deposit`, it will be the account ID of whoever relayed the message. So, this can only be used for reimbursing gas fees, but this functionality probably belongs to the connector itself, so no need to pass this to the recipient.

bridge-token-factory/src/lib.rs
```diff
@@ -196,13 +255,34 @@ impl BridgeTokenFactory {
...
+                MINT_GAS,
+            ),
+            Recipient::ContractId {
+                evm_address,
+                message,
+            } => ext_bridge_token::mint(
+                env::current_account_id(),
+                amount.into(),
+                &self.get_bridge_token_account_id(token),
+                env::attached_deposit() - required_deposit,
```
abacabadabacaba, line 273: Should we send the deposit to `mint` or to `ft_transfer_call`? The destination account may need to be created by `ft_transfer_call`.

bridge-token-factory/src/lib.rs
```diff
@@ -98,12 +102,67 @@ pub trait FungibleToken {
...
+            message,
+        }
+    } else {
+        Recipient::AccountId(recipient)
+    }
+}
+
+fn is_recipient_near_account_id(recipient: &String) -> bool {
+    !recipient.contains(':')
+}
```
abacabadabacaba, line 164: Is this function used anywhere?

bridge-token-factory/src/lib.rs
```diff
@@ -164,7 +223,7 @@ impl BridgeTokenFactory {
             proof_1,
             &env::current_account_id(),
             env::attached_deposit(),
-            FINISH_DEPOSIT_GAS + MINT_GAS,
+            FINISH_DEPOSIT_GAS + MINT_GAS + FT_TRANSFER_CALL_GAS,
```
abacabadabacaba, line 226: We probably only need to charge `FT_TRANSFER_CALL_GAS` if `ft_transfer_call` is going to be used.

abacabadabacaba (changes requested):

bridge-token-factory/src/lib.rs
mfornet: ft_transfer_call must receive exactly 1 yoctoNEAR. So target account in Aurora must exist before initiating this transaction.

bridge-token-factory/src/lib.rs
mfornet: https://github.com/aurora-is-near/rainbow-token-connector/issues/64

PR #71, opened 2021-07-29
feat: token metadata connector

0x3bfc:
- [x] ERC20MetadataLogger on Ethereum side
- [x] Update NEP-141 metadata on the NEAR side
- [ ] Update ERC20 metadata on Aurora side

Related issues:

- https://github.com/aurora-is-near/rainbow-bridge-frontend/issues/300
- https://github.com/aurora-is-near/rainbow-token-connector/issues/72


bridge-token-factory/src/lib.rs
```diff
@@ -110,6 +128,15 @@ pub trait ExtBridgeToken {
...
     ) -> PromiseOrValue<U128>;
+
+    fn set_metadata(
+        &mut self,
+        name: Option<String>,
+        symbol: Option<String>,
+        reference: Option<String>,
+        reference_hash: Option<Base64VecU8>,
+        decimals: Option<u8>,
+    );
```
sept-en, line 139: Need to add `icon: Option<String>,` field there for [NEP-148](https://nomicon.io/Standards/FungibleToken/Metadata.html) compliance

bridge-token-factory/src/lib.rs
```diff
@@ -92,6 +96,20 @@ pub trait ExtBridgeTokenFactory {
...
+        #[serializer(borsh)]
+        verification_success: bool,
+        #[serializer(borsh)] token: String,
+        #[serializer(borsh)] name: String,
+        #[serializer(borsh)] symbol: String,
+        #[serializer(borsh)] decimals: u8,
+        #[serializer(borsh)] timestamp: u64,
+        #[serializer(borsh)] proof: Proof,
+    ) -> Promise;
 }
```
sept-en, line 113: As the metadata could be updated either by calling either `set_metadata` directly or by verifying the proof from the `MetadataConnector`, I suggest renaming the method to something like:  `finish_updating_metadata_with_proof` to be explicit on its purpose and to eliminate possible misleading.

bridge-token/src/lib.rs
```diff
@@ -68,13 +78,14 @@ impl BridgeToken {
...
-        decimals.map(|decimals| self.decimals = decimals);
+        // Only controller of self can change the metadata
+        if self.assert_caller_self_or_factory() {
+            name.map(|name| self.name = name);
+            symbol.map(|symbol| self.symbol = symbol);
+            reference.map(|reference| self.reference = reference);
+            reference_hash.map(|reference_hash| self.reference_hash = reference_hash);
+            decimals.map(|decimals| self.decimals = decimals);
+        }
     }
```
sept-en, line 89: Add `icon` field here and in the function signature.

bridge-token-factory/src/lib.rs
```diff
@@ -224,6 +291,44 @@ impl BridgeTokenFactory {
...
+        env::log(format!("Finish updating metadata. Name:{} Symbol:{:?} Decimals:{:?} at : {:?}", name, symbol, decimals, timestamp).as_bytes());
+        let reference: Option<String> = Some(String::default());
+        let reference_hash: Option<Base64VecU8> = Some(Base64VecU8(vec![]));
+        // if timestamp > last updated at 
+        ext_bridge_token::set_metadata(
+            name.into(),
+            symbol.into(),
+            reference, 
+            reference_hash,
+            decimals.into(),
```
sept-en, line 325: Why do you overwrite token's `reference` and `reference_hash` with empty data? As these fields don't exist in the `MetadataConnector`'s `Log`, just set those to `None` so they won't be updated in `set_metadata()` method.

bridge-token-factory/src/lib.rs
```diff
@@ -224,6 +291,44 @@ impl BridgeTokenFactory {
...
+    ) {
+        assert_self();
+        assert!(verification_success, "Failed to verify the proof");
+        let required_deposit = self.record_proof(&proof);
+
+        assert!(
+            env::attached_deposit()
+                >= required_deposit + self.bridge_token_storage_deposit_required
+        );
+        env::log(format!("Finish updating metadata. Name:{} Symbol:{:?} Decimals:{:?} at : {:?}", name, symbol, decimals, timestamp).as_bytes());
```
sept-en, line 316: It's not clear what's `at` for without checking the sources, so either convert the timestamp to Human date or add the suffix telling that this is indeed a timestamp. IMO the second option is preferable as it doesn't require any additional computation.

bridge-token/src/lib.rs
```diff
@@ -60,6 +60,16 @@ impl BridgeToken {
         }
     }
 
+    pub fn assert_caller_self_or_factory(&mut self) -> bool {
+        if env::predecessor_account_id() == self.controller ||
+        env::predecessor_account_id() == env::current_account_id()
+        {
```
sept-en, line 66:
```suggestion
        if env::predecessor_account_id() == self.controller 
            || env::predecessor_account_id() == env::current_account_id() {
```

The indentation and stylistic are wrong.

sept-en (changes requested):
It'd be super nice to have this connector, thanks for adding this @0x3bfc !
At the first sight this LGTM!

mfornet (approved):

PR #73, opened 2021-08-18
BridgeToken: add `icon` metadata field. Update metadata

### LINKED ISSUE #72: `set_metadata` of bridge token doesn't support `icon` field
The `set_metadata` method now doesn't support icon field, which is necessary to update in many cases

https://github.com/aurora-is-near/rainbow-token-connector/blob/a45e23aed6d3566ea8f89da7d9527da931699fe6/bridge-token/src/lib.rs#L63-L78

We need to support `icon` field update. 

Please feel free to refer to the FT metadata Rust implementation if necessary: https://github.com/near/near-sdk-rs/blob/ddd9affe7dc9433d6e2abe454841d416d80cb43f/near-contract-standards/src/fungible_token/metadata.rs#L9-L17

sept-en:
Fixes #72


mfornet:
Steps to update the contracts:

1. Compile bridge-token and bridge-token-factory
2. Deploy factory in place (after this step all newly deployed tokens will use the new binary)
3. Pause deposits in factory (Not required to pause deployments)
4. Upgrade each individual bridge token
4.1 Deploy bridge token in place
4.2 Call migration function to update internal state
5. Unpause deposits

This will create a downtime for transferring tokens from Ethereum to NEAR between [3. and 5.)

bridge-token/src/lib.rs
```diff
@@ -67,14 +71,18 @@ impl BridgeToken {
...
+        assert!(self.controller_or_self());
+
         name.map(|name| self.name = name);
         symbol.map(|symbol| self.symbol = symbol);
         reference.map(|reference| self.reference = reference);
         reference_hash.map(|reference_hash| self.reference_hash = reference_hash);
         decimals.map(|decimals| self.decimals = decimals);
+        #[cfg(feature = "migrate_icon")]
+        icon.map(|icon| self.icon = Some(icon));
     }
```
sept-en, line 86: if we accept a function argument without the feature enabled, then we need to apply it without the feature as well. Or explicitly specify that we ignore that and don't allow making such a call with the provided icon. Otherwise, it's misleading

bridge-token/src/lib.rs
```diff
@@ -111,6 +119,12 @@ impl BridgeToken {
     pub fn account_storage_usage(&self) -> StorageUsage {
         self.token.account_storage_usage
     }
+
+    /// Return true if the caller is either self
+    pub fn controller_or_self(&self) -> bool {
+        let caller = env::predecessor_account_id();
+        caller == self.controller || caller == env::current_account_id()
+    }
 }
```
sept-en, line 128: I remember we discussed allowing changing the metadata not only for the token account itself and factory (which is the controller) but also for some external account like `controller.factory.bridge.near`. Is there any objection to not adding it?

bridge-token/src/lib.rs
```diff
@@ -134,3 +149,49 @@ impl FungibleTokenMetadataProvider for BridgeToken {
...
+}
+
+#[cfg(feature = "migrate_icon")]
+#[near_bindgen]
+impl BridgeToken {
+    /// Adding icon as suggested here: https://nomicon.io/Standards/FungibleToken/Metadata.html
+    /// This function can only be called from the factory or from the contract itself.
+    #[init(ignore_state)]
+    pub fn migrate_nep_148_add_icon() -> Self {
+        let old_state: BridgeTokenV0 = env::state_read().unwrap();
```
sept-en, line 192: Why do we `unwrap`? I think it's better to be explicit putting something like:
`env::state_read().expect("The state doesn't exist for BridgeTokenV0. Perhaps, migration is not required");`

bridge-token/src/lib.rs
mfornet: To avoid adding the controller on every single bridge token, I added the factory as the "controller" and will add a function to factory that will allow making this call from a controller.

bridge-token/src/lib.rs
mfornet: I tried doing this:

```rust
pub fn set_metadata(
        &mut self,
        name: Option<String>,
        symbol: Option<String>,
        reference: Option<String>,
        reference_hash: Option<Base64VecU8>,
        decimals: Option<u8>,
        #[cfg(feature = "migrate_icon")] icon: Option<String>,
    )
```

But it doesn't work

```
error[E0061]: this function takes 5 arguments but 6 arguments were supplied
  --> src/lib.rs:67:12
   |
67 |       pub fn set_metadata(
   |       -      ^^^^^^^^^^^^ expected 5 arguments
   |  _____|
   | |
68 | |         &mut self,
69 | |         name: Option<String>,
   | |         ----
70 | |         symbol: Option<String>,
   | |         ------
71 | |         reference: Option<String>,
   | |         ---------
72 | |         reference_hash: Option<Base64VecU8>,
   | |         --------------
73 | |         decimals: Option<u8>,
   | |         --------
74 | |         #[cfg(feature = "migrate_icon")] icon: Option<String>,
   | |                                          ---- supplied 6 arguments
75 | |     ) {
   | |_____- defined here

error: aborting due to previous error

For more information about this error, try `rustc --explain E0061`.
error: could not compile `bridge-token`.

To learn more, run the command again with --verbose.
make: *** [res/bridge_token_no_icon.wasm] Error 101
```

The other option I see is having two different functions, but didn't liked that approach. Are you suggesting to have the two functions, one for "migrate_icon" feature, and one without it? I think it is ok to have the function receiving the icon and not using it. The main reason for me to add this feature flag is for testing purposes.

bridge-token/src/lib.rs
sept-en: There's also `expect` method for the `Option`

bridge-token/src/lib.rs
sept-en: Ok, nice!

bridge-token/src/lib.rs
sept-en: Yes, I think two different functions could work but I don't like the approach, seems for me like a redundant overhead. Let's just add at least a log that if the icon is provided and `migrate_icon` feature is not enabled, let's print a log that the icon is being ignored.

mfornet: added log message.

bridge-token/src/lib.rs
mfornet: 👌

bridge-token-factory/src/lib.rs
```diff
@@ -378,6 +395,50 @@ impl BridgeTokenFactory {
...
+        env::storage_write(CONTROLLER_STORAGE_KEY, controller.as_bytes());
+    }
+
+    pub fn controller_or_self(&self) -> bool {
+        let caller = env::predecessor_account_id();
+        caller == env::current_account_id()
+            || self
+                .get_controller()
+                .map(|controller| controller == caller)
+                .unwrap_or_default()
```
sept-en, line 440: `.unwrap_or_default()` is dangerous here, because in case if default value will be true for some reason, this will be evaluated as `caller == env::current_account_id() || true` which will be `true`.

So I suggest using `.unwrap_or(false)`

bridge-token/src/lib.rs
```diff
@@ -189,7 +191,8 @@ impl BridgeToken {
     /// This function can only be called from the factory or from the contract itself.
     #[init(ignore_state)]
     pub fn migrate_nep_148_add_icon() -> Self {
-        let old_state: BridgeTokenV0 = env::state_read().unwrap();
+        let old_state: BridgeTokenV0 = env::state_read()
+            .expect("State is not compatible with BridgeTokenV0. Migration not applied.");
```
sept-en, line 195: Let's use either "Migration has not been applied" or "Migration is not applied"

bridge-token-factory/src/lib.rs
```diff
@@ -378,6 +395,50 @@ impl BridgeTokenFactory {
...
+            icon,
+            &self.get_bridge_token_account_id(address),
+            env::attached_deposit(),
+            env::prepaid_gas() - OUTER_SET_METADATA_GAS,
+        )
+    }
+
+    pub fn get_controller(&self) -> Option<AccountId> {
+        env::storage_read(CONTROLLER_STORAGE_KEY)
+            .map(|value| String::from_utf8(value).expect("Invalid controller account id"))
```
sept-en, line 425: WDYT, should we add another prefix for storage beside the `CONTROLLER_STORAGE_KEY`? Or are you considering that `a` in the `CONTROLLER_STORAGE_KEY` as that prefix?

bridge-token/src/lib.rs
```diff
@@ -83,6 +83,8 @@ impl BridgeToken {
         decimals.map(|decimals| self.decimals = decimals);
         #[cfg(feature = "migrate_icon")]
         icon.map(|icon| self.icon = Some(icon));
+        #[cfg(not(feature = "migrate_icon"))]
+        env::log("Icon is not supported".as_bytes())
```
sept-en, line 87: We should `log` this only if the icon is `Some` with something like: "Icon was provided, but it's not supported for the token"

bridge-token-factory/src/lib.rs
mfornet: Yep, I'm considering the `a` as that prefix. Didn't want to over-engineer there since we don't need it. But the idea is to use `a` as a prefix for custom variables to avoid potential conflicts in the future.

sept-en: Ok, I agree!

mfornet:
Migrate tokens using [bridge-cli](https://github.com/aurora-is-near/bridge-cli). Steps to do it:

1. Install bridge-cli using `npm i -g rainbow-bridge`
2. Check it is installed properly `bridge --version`
3. Check list of all available bridge `bridge list`
4. Select bridge `bridge use ropsten`
5. Pause the factory `bridge factory pause --deposit`.
6. Check paused state of the factory with: `bridge factory pause --status`
7. Fetch list with all deployed tokens: `bridge tokens list > ropsten.txt`
8. Compile new contract to be deployed from this branch in this repository. It can be found in (res/bridge_token.wasm)
9. Run migration `bridge tools migrate-icons ropsten.txt res/bridge_token.wasm`
10. Unpause the factory: `bridge factory unpause --deposit`

bridge-token-factory/src/lib.rs
```diff
@@ -370,9 +526,74 @@ impl BridgeTokenFactory {
...
+        UnorderedMap::new(TOKEN_TIMESTAMP_MAP_PREFIX.to_vec())
+    }
+
+    /// Factory Controller. Controller has extra privileges inside this contract.
+    pub fn controller(&self) -> Option<AccountId> {
+        env::storage_read(CONTROLLER_STORAGE_KEY)
+            .map(|value| String::from_utf8(value).expect("Invalid controller account id"))
+    }
+
+    pub fn use_controller(&mut self, controller: AccountId) {
```
sept-en, line 570:
```suggestion
    pub fn set_controller(&mut self, controller: AccountId) {
```

bridge-token-factory/src/lib.rs
```diff
@@ -370,9 +526,74 @@ impl BridgeTokenFactory {
...
+    }
+
+    /// Ethereum Metadata Emitter. This is the address where the contract that emits metadata from tokens
+    /// on ethereum is deployed. Address is encoded as hex.
+    pub fn metadata_emitter(&self) -> Option<String> {
+        env::storage_read(METADATA_EMITTER_STORAGE_KEY)
+            .map(|value| String::from_utf8(value).expect("Invalid metadata emitter address"))
+    }
+
+    pub fn use_metadata_emitter(&mut self, metadata_emitter: String) {
```
sept-en, line 592:
```suggestion
    pub fn set_metadata_connector(&mut self, metadata_connector: String) {
```

bridge-token-factory/src/lib.rs
```diff
@@ -370,9 +526,74 @@ impl BridgeTokenFactory {
...
+        assert!(self.controller_or_self());
+        validate_eth_address(metadata_emitter.clone());
+        env::storage_write(METADATA_EMITTER_STORAGE_KEY, metadata_emitter.as_bytes());
+    }
```
sept-en, line 596: Please use `metadata_connector` instead of `metadata_emitter` everywhere.

bridge-token-factory/src/log_metadata_event.rs
```diff
@@ -0,0 +1,101 @@
...
+            metadata_emitter: [0u8; 20],
+            token: "6b175474e89094c44da98b954eedeac495271d0f".to_string(),
+            name: "TEST".to_string(),
+            symbol: "TST".to_string(),
+            decimals: 18,
+            timestamp: 13042194,
+        };
+        let data = event_data.to_log_entry_data();
+        let result = TokenMetadataEvent::from_log_entry_data(&data);
+        // assert_eq!(result, event_data);
```
sept-en, line 99: Why the test is not active?

bridge-token/src/lib.rs
```diff
@@ -57,6 +59,18 @@ impl BridgeToken {
...
+    }
+
+    pub fn assert_caller_self_or_factory(&mut self) -> bool {
+        if env::predecessor_account_id() == self.controller
+            || env::predecessor_account_id() == env::current_account_id()
+        {
+            return true;
+        } else {
+            return false;
         }
```
sept-en, line 74: It doesn't assert anything and duplicate `controller_or_self()` function logic.

metadata-connector/hardhat.config.js
```diff
@@ -0,0 +1,24 @@
...
+  //   ropsten: {
+  //     url: `https://ropsten.infura.io/v3/${process.env.INFURA_API_KEY}`,
+  //     accounts: [`0x${AURORA_PRIVATE_KEY}`],
+  //     chainId: 3,
+  //     live: true,
+  //     gasPrice: 50000000000,
+  //     gasMultiplier: 2,
+  //   },
+  // }
+};
```
sept-en, line 24: Please fix commented out code.

metadata-connector/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "name": "metadata_connector",
+  "version": "1.0.0",
+  "main": "index.js",
+  "license": "MIT",
+  "scripts": {
+    "clean": "rm -rf ./artifacts/ ./cache/",
+    "compile": "npx hardhat compile",
+    "test": "npx hardhat test",
+    "deploy:aurora": "yarn clean && yarn compile && npx hardhat run scripts/deploy.js --network aurora",
+    "deploy:ropsten": "yarn clean && yarn compile && npx hardhat run scripts/deploy.js --network ropsten"
```
sept-en, line 11: Please add `deploy:goerli` and `deploy:mainnet` commands

metadata-connector/test/ERC20MetadataLogger.test.js
```diff
@@ -0,0 +1,22 @@
...
+    const { events } = await tx.wait()
+    const args = events.find(({ event }) => event === 'Log').args
+    expect(args.erc20).to.equal(sampleERC20.address)
+    expect(args.name).to.equal("SampleERC20")
+    expect(args.symbol).to.equal("ERC")
+    expect(args.decimals).to.equal(18)
+  });
+  it("Should revert if in case of non-erc20 compliant", async function () {
+    //TBD
+  })
```
sept-en, line 21: This is not implemented.

metadata-connector/package.json
mfornet: removed this instead, added the function to deploy tokens from bridge-cli

sept-en (commented):
LGTM!

mfornet (approved):

PR #74, opened 2021-08-27
Migrate CI pipeline to GitHub actions

strokovok:
Seems like everything works exactly as intended

PR #78, opened 2021-11-10
CI fixes

sept-en:
N.B. I don't have access to `buildkite` anymore. So this PR just fixes the obvious CI check, no need to care for `buildkite` because of upcoming #74


karim-en (approved):
Nothing to review, as I see it is just about updating binaries by rebuilding the contracts.

PR #82, opened 2021-12-06
check the proof by Tx height

### LINKED ISSUE #56: Fix min block acceptance height threshold
In order to make upgrades to the Locker contract without requiring to move all used proofs from one contract to another, a height threshold is used, where only proofs that were created after this threshold (with higher heights) can be used, and others will be considered invalid. 

For this end we have [this check](https://github.com/aurora-is-near/rainbow-token-connector/blob/master/erc20-connector/contracts/Locker.sol#L36). 

In this case `proofBlockHeight` refers to the height of the block used to build the proof, and not the height of the block where the transaction was included. Instead of using `proofBlockHeight` we should use height from the proof ( `fullOutcomeProof...height`).

There is no way to hack this bug at the moment, but it is critical to fix this in case the Locker is going to be upgraded.

idirall22:
- Check if the proof is not from an ancient block. The check is done using by comparing the Tx finality height and the minBlockAcceptanceHeight.

Fixes #56

### DIFF
--- a/erc20-connector/test/TokenLocker.js
+++ b/erc20-connector/test/TokenLocker.js
@@ -208,6 +208,23 @@ contract('TokenLocker', function ([adminAddress, addr1, addr2]) {
         });
     });
 
+    it.only('can not unlock, proof is invalid', async function () {
+        let proof = require('./proof_template.json');
+        proof.outcome_proof.outcome.status.SuccessValue = serialize(SCHEMA, 'Unlock', {
+            flag: 0,
+            amount: toWei('1'),
+            token: hexToBytes(token.address),
+            recipient: hexToBytes(addr1),
+        }).toString('base64');
+        const lockerBalance = await token.balanceOf(locker.address);
+        const receiverBalance = await token.balanceOf(addr1);
+
+        await expectRevert(
+            locker.unlockToken(borshifyOutcomeProof(proof), 1099),
+            "Proof is from the ancient block"
+            )
+    });
+
     /*it('deposit & withdraw', async function() {
         assert(await locker.isBridgeToken(btoken.address));
         expect(await locker.ethToNearToken(btoken.address)).equal(nearToken);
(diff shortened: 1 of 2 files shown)


mfornet (approved):
LGTM.

erc20-connector/test/TokenLocker.js
```diff
@@ -208,6 +208,23 @@ contract('TokenLocker', function ([adminAddress, addr1, addr2]) {
         });
     });
 
+    it.only('can not unlock, proof is invalid', async function () {
```
sept-en, line 211:
```suggestion
    it.only('can not unlock, proof is from the ancient block', async function () {
```

erc20-connector/test/TokenLocker.js
```diff
@@ -208,6 +208,23 @@ contract('TokenLocker', function ([adminAddress, addr1, addr2]) {
...
+            token: hexToBytes(token.address),
+            recipient: hexToBytes(addr1),
+        }).toString('base64');
+        const lockerBalance = await token.balanceOf(locker.address);
+        const receiverBalance = await token.balanceOf(addr1);
+
+        await expectRevert(
+            locker.unlockToken(borshifyOutcomeProof(proof), 1099),
+            "Proof is from the ancient block"
+            )
```
sept-en, line 225:
```suggestion
        );
```

sept-en (changes requested):

erc20-connector/test/TokenLocker.js
sept-en: to be fixed in a follow-up PR.

erc20-connector/test/TokenLocker.js
sept-en: to be fixed in a follow-up PR.

sept-en (approved):

sept-en:
Due to the lack of access to the source branch, merge conflicts will be resolved and thus merged to `maser` in the following-up PR.

PR #87, opened 2021-12-31
Borsh: check if the decoding is finished

sept-en:


erc20-connector/contracts/ERC20Locker.sol
```diff
@@ -76,6 +76,7 @@ contract ERC20Locker is Locker, AdminControlled {
         result.token = address(uint160(token));
         bytes20 recipient = borshData.decodeBytes20();
         result.recipient = address(uint160(recipient));
+        require(borshData.finished(), "Parse error: EOI expected");
```
karim-en, line 79: Could you please explain how finish releated to borshData.done() which is described in the issue?

sept-en: @karim-en this is actually the same function. `finished()` was replaced by `done()` in future versions, but we will need to bump up used `rainbow-bridge` package version later.

karim-en (approved):

mfornet (approved):

PR #93, opened 2022-02-02
Solidity: naming consistency fixes

sept-en:
The rainbow-bridge contracts don't use trailing underscores for the public variables' names, so we need to keep consistency across all connectors.


mfornet (approved):
Thanks, I prefer the new convention rather than the old one.

One suggestion, should we add underscore at the end per [Style Guide](https://docs.soliditylang.org/en/v0.8.10/style-guide.html?#avoiding-naming-collisions). Not blocking.

karim-en (approved):

sept-en:
@mfornet we don't use those trailing underscores in most other contracts so the initial goal is to get consistency. And also the drawback I think is that it's not that convenient using it from API/Etherscan.
However, I think that we may consider changing everything to underscores too, WDYT?

Issue #133, opened 2022-08-12
FT connector: improve pausability for Solidity contracts


sept-en:
Currently, FT-connector's contracts use the `PausableUpgradable` contract implemented by OZ to pause functions. The drawback is that this way all the functions marked with the modifier will be paused/unpaused. We need to have the ability to pause only some specific functions. For that purpose, we can use the `AdminControlled` contract, the same way as it's used for `ERC20Locker`.

PR #152, opened 2022-08-23
FT connector: improve pausability

karim-en:
Fix https://github.com/aurora-is-near/rainbow-token-connector/issues/133


erc20-bridge-token/contracts/SelectivePausableUpgradable.sol
```diff
@@ -0,0 +1,110 @@
...
+abstract contract SelectivePausableUpgradable is Initializable, ContextUpgradeable {
+    /**
+     * @dev Emitted when the pause is triggered by `account`.
+     */
+    event Paused(address account);
+
+    /**
+     * @dev Emitted when the pause is lifted by `account`.
+     */
+    event Unpaused(address account);
```
sept-en, line 26: Please pass flags into the event

sept-en (changes requested):
Currently, `unpause` sets the flags to 0, however, `pause` also allows unpausing by setting the `flags` as `0`. We also need to have a separate roles for pausing/unpausing. I think, to achieve this we might replace the current `PAUSE_ROLE` with `ADMIN_PAUSE_ROLE` and also add a couple of functions that will allow pausing only a specific method (hardcoded in the function) with a given role: i.e. `PAUSE_DEPOSIT_ROLE`, `PAUSE_WITHDRAWAL_ROLE`, etc.

erc20-bridge-token/contracts/SelectivePausableUpgradable.sol
```diff
@@ -0,0 +1,104 @@
...
+ * the functions of your contract. Note that they will not be pausable by
+ * simply including this module, only once the modifiers are put in place.
+ */
+abstract contract SelectivePausableUpgradable is Initializable, ContextUpgradeable {
+    /**
+     * @dev Emitted when the pause is triggered by `account`.
+     */
+    event Paused(address account, uint flags);
+
+    bool private _paused; // Deprecated!
```
sept-en, line 23: I think we should remove the deprecated field? If I get it correctly, this is related to the layout, then we need to have an appropriate name and description.

karim-en: Ok, removed since we haven't deployed it yet on mainnet

erc20-bridge-token/contracts/SelectivePausableUpgradable.sol
karim-en: 788d0a9d350c66faa99cfd50b585391c8bb7bf13

erc20-bridge-token/contracts/SelectivePausableUpgradable.sol
karim-en: e16126cedc28081154f50cd4576fc46b0b1634fc

sept-en (approved):

PR #166, opened 2022-09-08
FT & ERC20 connectors: fix parsing recipient with custom message

### LINKED ISSUE #160: FT connector: add an option to disable whitelist mode
In the FT connector, the whitelist mode is enabled by default. However, there are no controls allowing to switch it off. The appropriate interface should be added.

### LINKED ISSUE #162: FT connector: `BridgeToken` pause/unpause interface
The role of pauser and admin is granted to `BridgeTokenFactory` during the initialization of the token. However, there is no interface in `BridgeTokenFactory` to pause or unpause the Bridge token. There is also no interface in `BridgeTokenFactory` to grant a pauser role for other accounts. Thus, Bridge tokens can’t be paused currently.

### LINKED ISSUE #163: FT connector: `token-locker` is unable to parse custom messages
The contract has the withdraw functionality, the flow of which is to check the given proof and give money to the recipient user account. ETH side can add additional messages into the event (data will be part of the proof), and it will be contained in the same data field where the recipient account is stored. In regular cases, these strings should be separated with “:”, and the NEAR side should parse it well. In the current implementation, when proof from the byte array is parsed, it is casted to NEAR `AccountId` type where the symbol ‘:’ is not allowed. So in case the contract receives proof with this additional message - the contract will crash while trying to make `AccountId` type from this string.

karim-en:
Fix #163
Fix #160


sept-en (changes requested):
Please add a test utilizing custom message protocol

karim-en:
> Please add a test utilizing custom message protocol

Done

sept-en (approved):

PR #174, opened 2022-10-05
Token-locker: comments and review

olga24912:


token-locker/src/lib.rs
```diff
@@ -326,6 +363,7 @@ impl Contract {
         required_deposit
     }
 
+    /// Update Ethereum Address of the Bridge Token Factory
     #[private]
     pub fn update_factory_address(&mut self, factory_address: String) {
         self.eth_factory_address = validate_eth_address(factory_address);
```
olga24912, line 384: I think `validate_eth_address` is not a clear name, it is more like a constructor for EthAddress from String. I think the name `eth_address_from_string` is more clear. I think it is Ok to do some checks in the constructor. I expect that the function `validate_eth_address` return is this address is valid or not.

token-locker/src/lib.rs
```diff
@@ -68,54 +73,80 @@ pub struct Contract {
     pub is_whitelist_mode_enabled: bool,
```
olga24912, line 73: The Ethereum Factory also have a whitelist. Are these whitelists consistent with each other? If yes, how is this achieved?

token-locker/src/lib.rs
```diff
@@ -300,7 +350,7 @@ impl Contract {
         }
     }
 
-    /// Checks whether the provided proof is already used.
+    /// Checks whether the provided proof of Ethereum `burn` event is already used for unlocking tokens
     pub fn is_used_proof(&self, #[serializer(borsh)] proof: Proof) -> bool {
```
olga24912, line 356: Where this function is used? Is it just a part of public API?

token-locker/src/lib.rs
```diff
@@ -274,8 +324,10 @@ impl Contract {
         #[serializer(borsh)] proof: Proof,
     ) -> Promise {
         assert!(verification_success, "Failed to verify the proof");
-        let required_deposit = self.record_proof(&proof);
 
+        // The deposit required to cover the cost
+        // of additional storage of used proof
+        let required_deposit = self.record_proof(&proof);
         assert!(env::attached_deposit() >= required_deposit);
```
olga24912, line 331: It also should be an extra 2 YOCTO which we used for ft_transfer_call. And maybe we want to return extra money back if the deposit is bigger than required.

token-locker/src/lib.rs
```diff
@@ -274,8 +324,10 @@ impl Contract {
         #[serializer(borsh)] proof: Proof,
     ) -> Promise {
         assert!(verification_success, "Failed to verify the proof");
-        let required_deposit = self.record_proof(&proof);
 
+        // The deposit required to cover the cost
+        // of additional storage of used proof
+        let required_deposit = self.record_proof(&proof);
```
olga24912, line 330: What if we get an error during ft_transfer_call? In that case, we will record that proof is already used, but we don't unlock tokens.

erc20-bridge-token/contracts/ProofConsumer.sol
```diff
@@ -81,17 +81,17 @@ contract ProofConsumer is Ownable, IProofConsumer {
                     .outcome
                     .executor_id
             ) == keccak256(nearTokenLocker),
```
olga24912, line 83: Maybe it is better to do all these checks at the begging of the function. I think we shouldn't store it in the `usedProofs` incorrect proofs.

PR #178, opened 2022-11-04
Near-plugins integration

karim-en:
Integrate near plugins to the `bridge token factory` and `ft token locker`.

Details:
- The near plugins weren't integrated into the bridge token because they are fully controlled by the factory contract.
- Added `upgrade_bridge_token(&self, address: String)` to upgrade the bridge token from the factory.
- Added `set_paused_withdraw(&mut self, address: String, paused: bool)` to pause the token from the factory.
- Added `upgrade_and_migrate` to the bridge token contract, which applies the upgrade and calls `migrate(from_version: u32)` where the `from_version` is passed as an argument to be able to migrate from version `A` to `C` if the upgrade to `B` was missed by mistake.


olga24912 (changes requested):

bridge-token/src/lib.rs
```diff
@@ -56,7 +59,14 @@ impl BridgeToken {
             paused: Mask::default(),
             #[cfg(feature = "migrate_icon")]
             icon: None,
-        }
+        };
+
+        ::near_sdk::env::storage_write(
```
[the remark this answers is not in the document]
karim-en: This will not work because the contract is deployed by the factory contract and the `owner_set` doesn't allow to set the owner if the predecessor account is not the current account

token-locker/src/lib.rs
```diff
@@ -178,8 +184,8 @@ impl Contract {
     /// Withdraw funds from NEAR Token Locker.
     /// Receives proof of burning tokens on the other side. Validates it and releases funds.
     #[payable]
+    #[pause(except(owner, self))]
     pub fn withdraw(&mut self, #[serializer(borsh)] proof: Proof) -> Promise {
-        self.check_not_paused(PAUSE_DEPOSIT);
         let event = EthUnlockedEvent::from_log_entry_data(&proof.log_entry_data);
```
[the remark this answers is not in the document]
karim-en: The "PAUSE_DEPOSIT" naming is not correct, it was supposed to be "PAUSE_WITHDRAW"

olga24912 (approved):

UrAvgDeveloper (changes requested):
Can we add access control plugin in bridge token contract as well? this way we don't need `admin-controlled` as well maybe and can configure access according to functions

bridge-token/src/lib.rs
```diff
@@ -98,8 +111,7 @@ impl BridgeToken {
 
     #[payable]
     pub fn withdraw(&mut self, amount: U128, recipient: String) -> Promise {
-        self.check_not_paused(PAUSE_WITHDRAW);
-
+        require!(!self.is_paused());
```
olga24912, line 114: Why we don't use the near plugins for this contract?

karim-en: The token contract can be controlled just by the `Factory` contract, it was possible to do that when plugins like the Pausable plugin were implemented based on Ownable, but now the near plugins use ACL everywhere which complicates the maintenance and the implementation.

olga24912 (approved):

bridge-token/src/lib.rs
```diff
@@ -44,7 +46,17 @@ pub trait ExtBridgeTokenFactory {
...
+        let current_account_id = env::current_account_id();
+        let (_eth_address, factory_account) = current_account_id
+            .as_str()
+            .split_once(".")
+            .unwrap_or_else(|| env::panic_str("Invalid token address"));
+
+        require!(
+            env::predecessor_account_id().as_str() == factory_account,
+            "Only the factory account can init this contract"
+        );
```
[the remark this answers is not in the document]
karim-en: Yes, only the factory can create it, but anyone can call new/init because it is a public method.
btw, I have an idea of how this could be fixed in another way, by call the deeply and new in batch transaction

bridge-token/src/lib.rs
```diff
@@ -177,17 +215,19 @@ impl From<BridgeTokenV0> for BridgeToken {
     }
 }
 
-#[cfg(feature = "migrate_icon")]
 #[near_bindgen]
 impl BridgeToken {
-    /// Adding icon as suggested here: https://nomicon.io/Standards/FungibleToken/Metadata.html
     /// This function can only be called from the factory or from the contract itself.
```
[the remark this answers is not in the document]
karim-en: Due to the check `assert!(new_state.controller_or_self());`

bridge-token-factory/src/lib.rs
```diff
@@ -486,39 +528,26 @@ impl BridgeTokenFactory {
...
-    }
-
     /// Ethereum Metadata Connector. This is the address where the contract that emits metadata from tokens
     /// on ethereum is deployed. Address is encoded as hex.
     pub fn metadata_connector(&self) -> Option<String> {
         env::storage_read(METADATA_CONNECTOR_ETH_ADDRESS_STORAGE_KEY)
             .map(|value| String::from_utf8(value).expect("Invalid metadata connector address"))
     }
 
+    #[access_control_any(roles(Role::DAO))]
```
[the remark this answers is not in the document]
karim-en: This function is sensitive, so only DAO should be able to apply changes like this.

bridge-token/src/lib.rs
```diff
@@ -24,11 +27,10 @@ pub struct BridgeToken {
     reference_hash: Base64VecU8,
     decimals: u8,
     paused: Mask,
-    #[cfg(feature = "migrate_icon")]
     icon: Option<String>,
 }
 
-const PAUSE_WITHDRAW: Mask = 1 << 0;
+pub type Mask = u128;
```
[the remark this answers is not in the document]
karim-en: e266cb8f401cd90d64381861fe0f35c87ea1bf0c

bridge-token-factory/src/lib.rs
```diff
@@ -414,6 +439,23 @@ impl BridgeTokenFactory {
             )
     }
 
+    #[access_control_any(roles(Role::DAO, Role::UpgradableManager))]
+    pub fn upgrade_bridge_token(&self, address: String) -> Promise {
```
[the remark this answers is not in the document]
karim-en: e266cb8f401cd90d64381861fe0f35c87ea1bf0c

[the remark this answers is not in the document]
olga24912: https://github.com/aurora-is-near/rainbow-token-connector/pull/178/commits/d22458de60069cc444e07f7e09cd510b5c39f474

bridge-token/src/lib.rs
```diff
@@ -120,6 +125,45 @@ impl BridgeToken {
...
+        // Call promise to migrate the state.
+        // Batched together to fail upgrade if migration fails.
+        env::promise_batch_action_function_call(
+            promise_id,
+            "migrate",
+            &json!({ "from_version": CURRENT_STATE_VERSION })
+                .to_string()
+                .into_bytes(),
+            NO_DEPOSIT,
+            env::prepaid_gas() - env::used_gas() - UPGRADE_GAS_LEFTOVER,
```
olga24912, line 159: What if we provide to small amount of gas and deploy will be successful, but migrate don't. Maybe we should add extra check for gas amount?

bridge-token/src/lib.rs
```diff
@@ -120,6 +125,45 @@ impl BridgeToken {
...
+            self.controller_or_self(),
+            "Only the controller or self can update the code"
+        );
+
+        // Receive the code directly from the input to avoid the
+        // GAS overhead of deserializing parameters
+        let code = env::input().unwrap_or_else(|| panic!("ERR_NO_INPUT"));
+        // Deploy the contract code.
+        let promise_id = env::promise_batch_create(&env::current_account_id());
+        env::promise_batch_action_deploy_contract(promise_id, &code);
```
olga24912, line 149: Does the specification of how much gas it used exist?

bridge-token-factory/src/lib.rs
```diff
@@ -404,7 +440,6 @@ impl BridgeTokenFactory {
         Promise::new(bridge_token_account_id.parse().unwrap())
             .create_account()
             .transfer(BRIDGE_TOKEN_INIT_BALANCE)
-            .add_full_access_key(self.owner_pk.clone())
```
olga24912, line 407: How we use the full access key for owner before?

olga24912 (approved):

bridge-token/src/lib.rs
karim-en: It is batch call, so the deploy should be reverted if the migration failed

bridge-token/src/lib.rs
karim-en: It depends on the code size, here is more details 
https://docs.near.org/concepts/basics/transactions/gas-advanced

bridge-token-factory/src/lib.rs
karim-en: This key was added to the token contract to be able to upgrade it and call private methods.

karim-en: This key was added to the token contract to be able to upgrade it and call private methods.
Now we are going to remove this PK access, so the token should be controlled just by the factory contract, and the factory contract will be controlled by DAO.

olga24912 (approved):

PR #213, opened 2023-05-02
Fee integration

UrAvgDeveloper:
## Description

This PR contains Fee Implementation for Deposit and Withdrawal flows for tokens transfer using Rainbow Bridge (Business requirement).

Specification:
- Deposit and withdrawal can have separate fee percentages.
- The fee collected is locked in the same contract that the owner could claim using the function `claim_fee`.
- All the fee setters are only callable by the accounts that have the role `Role::FeeSetter`.
- The `fee_amount` has lower and upper bounds.
- If the the calculated `fee_amount` is higher than the `withdraw/deposit amount` then charge the full amount as a fee and don't transfer anything to the recipient.
- The fee is set in 6 decimal precision `10% = 0.1 * 10e6 = 100_000`
- It is possible to set a particular fee for silos on deposit and withdrawal. The silo is just a NEAR account, so as a side effect, it is possible to set a specific fee to any NEAR account. 
- It is possible not just to set a particular fee for a particular token for the silo, but also a default fee for this silo.

Added public function:
```
// Deposit
set_deposit_fee(token: EthAddressHex, fee_percentage: U128, lower_bound: Option<U128>, upper_bound: Option<U128>);
unset_deposit_fee(&mut self, token: EthAddressHex);
set_deposit_fee_per_silo(silo_account_id: AccountId, token: Option<EthAddressHex>, fee_percentage: U128, lower_bound: Option<U128>, upper_bound: Option<U128>);
unset_deposit_fee_per_silo(silo_account_id: AccountId, token: Option<EthAddressHex>);

// Withdraw
set_withdraw_fee(token: EthAddressHex, fee_percentage: U128, lower_bound: Option<U128>, upper_bound: Option<U128>);
unset_withdraw_fee(&mut self, token: EthAddressHex);
set_withdraw_fee_per_silo(silo_account_id: AccountId, token: Option<EthAddressHex>, fee_percentage: U128, lower_bound: Option<U128>, upper_bound: Option<U128>);
unset_withdraw_fee_per_silo(silo_account_id: AccountId, token: Option<EthAddressHex>);

// Getters
get_deposit_fee(&self, token: &EthAddressHex) -> Option<Fee>
get_withdraw_fee(&self, token: &EthAddressHex) -> Option<Fee>
get_withdraw_fee_per_silo(&self, silo_account_id: AccountId, token: Option<EthAddressHex>) -> Option<Fee>
get_deposit_fee_per_silo(&self, silo_account_id: AccountId, token: Option<EthAddressHex>) -> Option<Fee>

// Claim
claim_fee(&mut self, token: AccountId, amount: Balance) -> Promise

// Calculate 
calculate_deposit_fee_amount(&self, token: &EthAddressHex, amount: U128, target: Option<AccountId>) -> U128
calculate_withdraw_fee_amount(&self, token: &EthAddressHex, amount: U128, withdrawer: &AccountId) -> U128
```

Migration: 
The migration is needed due to the added fields `deposit_fee, withdraw_fee, withdraw_fee_per_silo, deposit_fee_per_silo` to the contract structure. To migrate the storage we need to call:
`pub fn migrate() -> Self`


bridge-token-factory/src/lib.rs
```diff
@@ -711,29 +1119,6 @@ mod tests {
...
-            predecessor_account_id: format!("{}.{}", token_locker(), bridge_token_factory())
-        );
-
-        let address = validate_eth_address(token_locker());
-        assert_eq!(
-            contract.finish_withdraw(1_000, token_locker()),
-            result_types::Withdraw::new(1_000, address, address)
-        );
-    }
-
```
sept-en, line 736: Why do you delete the test? It should still pass by configuring the zero fess.

bridge-token-factory/tests/token_transfer.rs
```diff
@@ -125,9 +142,28 @@ fn run_view_function(
...
+    } else {
+        return fee_amount;
+    }
+}
+
 #[test]
-fn test_eth_token_transfer() {
-    let (alice, factory, worker) = create_contract();
+fn test_token_transfer_with_deposit_and_withdraw_fee() {
+    let (alice, factory, _, worker) = create_contract();
```
sept-en, line 166: Please don't remove the already existing test. Fee functionality is optional and thus should have all old tests to pass without any issues. If there's a strong reason for it not to pass - please describe it accordingly.

bridge-token-factory/src/lib.rs
```diff
@@ -61,6 +61,9 @@ const METADATA_CONNECTOR_ETH_ADDRESS_STORAGE_KEY: &[u8] = b"aM_CONNECTOR";
 /// The prefix is made specially short since it becomes more expensive with larger prefixes.
 const TOKEN_TIMESTAMP_MAP_PREFIX: &[u8] = b"aTT";
 
+const FEE_DECIMAL_PRECISION: u128 = 1000000;
+const AURORA_ID: &str = "aurora";
```
sept-en, line 65: This shouldn't be fixed, as this could be an Aurora silo.

bridge-token-factory/src/lib.rs
```diff
@@ -75,6 +78,48 @@ pub enum Role {
...
+    eth_to_aurora: u128,
+}
+
+#[derive(
+    Default, BorshDeserialize, BorshSerialize, Debug, Clone, Serialize, Deserialize, PartialEq,
+)]
+pub struct WithdrawFeePercentage {
+    near_to_eth: u128,
+    aurora_to_eth: u128,
+}
```
sept-en, line 115: Why do you need regular `Serialize` and `Deserialize`?

sept-en (changes requested):
Thanks for the PR! I made a first shallow review and left some comments.

Besides that, please note that the scope of PR is quite big, so please make sure that all the changes are properly documented and that PR's description is adjusted accordingly. For example, the description should contain information on:
* how the fee is implemented;
* what the behavior would be;
* what the roles are;
* how to collect the fee;
* what are the new dependencies;
* whether this PR will require a storage migration;
* etc.

bridge-token-factory/src/lib.rs
UrAvgDeveloper: Can you please guide me, what predecessor account id should be when withdrawing from AURORA?

sept-en: It should be flexible - it should be allowed to be configurable. Also, maybe it makes sense to allow subsets of accounts, e.g., the admin configures the address of Aurora as `aurora` and sets the flag to allow subset silos. This will mean that all the accounts in the domain will be treated as Aurora accounts, e.g., `silo1.aurora`, `some_other_silo.aurora`, etc. However, this will definitely need to have an exception for Aurora XCC accounts on NEAR (the ones containing in the domain name an EVM address - 20bytes + `.aurora`)

bridge-token-factory/tests/token_transfer.rs
UrAvgDeveloper: This was done accidentally, added it back :)

bridge-token-factory/src/lib.rs
UrAvgDeveloper: The build fails if this is not included :(

bridge-token-factory/src/lib.rs
UrAvgDeveloper: fixed in 83383a048ac4ee32c5548ed19fec1a703a91a0dc

bridge-token-factory/src/lib.rs
UrAvgDeveloper: added in 7684b501ede66a8c1e151cfc89af0d486310f25f

karim-en: It is still hardcoded, please make it configurable.

karim-en: Also, I would recommend holding a map of allowed engine accounts like in the [aurora-eth-connector](https://github.com/aurora-is-near/aurora-eth-connector/blob/master/eth-connector/src/lib.rs#L57)
`known_engine_accounts`

bridge-token-factory/src/lib.rs
```diff
@@ -107,6 +160,16 @@ pub struct BridgeTokenFactory {
...
+    /// Fee amount bound for deposit.
+    pub deposit_fee_bound: UnorderedMap<String, DepositTokenBounds>,
+    /// Fee amount bound for withdraw
+    pub withdraw_fee_bound: UnorderedMap<String, WithdrawTokenBounds>,
+    /// Fee percentage of each token for deposit
+    pub deposit_fee_percentage: UnorderedMap<String, DepositFeePercentage>,
+    /// Fee percentage of each token for withdraw
+    pub withdraw_fee_percentage: UnorderedMap<String, WithdrawFeePercentage>,
+    /// Withdraw fee percentage for each different silos for different tokens
+    pub withdraw_fee_percentage_per_silo: UnorderedMap<AccountId, SiloTokenFeeAmount>,
```
karim-en, line 172: You need to implement a migration since new fields were added.
You can also remove the `#[deprecated] paused: Mask` field, it was marked as `deprecated` just to avoid migration.

bridge-token-factory/src/lib.rs
```diff
@@ -159,13 +222,36 @@ pub trait ExtBridgeToken {
...
 
 pub fn assert_self() {
     assert_eq!(env::predecessor_account_id(), env::current_account_id());
 }
 
+/*
+Checks whether the `withdrawer` is an aurora-chain account
+returns true iff `withdrawer` is an aurora account.
+ */
+pub fn is_aurora_engine_account(withdrawer: AccountId) -> bool {
```
karim-en, line 240: It is better to store the known engine accounts like in the [aurora-eth-connector](https://github.com/aurora-is-near/aurora-eth-connector/blob/master/eth-connector/src/lib.rs#L57)
known_engine_accounts

karim-en: And probably this can be combined with `withdraw_fee_percentage_per_silo`, so one map should be enough

karim-en (changes requested):

bridge-token-factory/src/lib.rs
```diff
@@ -78,6 +81,56 @@ pub enum Role {
...
+    Default, BorshDeserialize, BorshSerialize, Debug, Clone, Serialize, Deserialize, PartialEq,
+)]
+pub struct DepositFeePercentage {
+    eth_to_near: u128,
+    eth_to_aurora: u128,
+}
+
+#[derive(
+    Default, BorshDeserialize, BorshSerialize, Debug, Clone, Serialize, Deserialize, PartialEq,
+)]
```
karim-en, line 114: Please replace `U128` with `u128`, because those types are Serialize/Deserialize into json that doesn't support u128 integers.

bridge-token-factory/src/lib.rs
```diff
@@ -342,6 +434,212 @@ impl BridgeTokenFactory {
...
+
+    //this should be added as per: 10% -> 0.1 = 0.1*10^6
+    #[access_control_any(roles(Role::FeeSetter))]
+    pub fn set_withdraw_fee_percentage(
+        &mut self,
+        token: &String,
+        near_to_eth: u128,
+        aurora_to_eth: u128,
+    ) -> WithdrawFeePercentage {
+        let _ = validate_eth_address(token.clone());
```
karim-en, line 548: I think it will be good to implement ser/des for EthAddress type and pass it instead of a string.
Here is the [implementation](https://github.com/aurora-is-near/fast-bridge-common/blob/master/src/lib.rs#L13) , you can move it to the `bridge-common` crate.

bridge-token-factory/src/lib.rs
```diff
@@ -342,6 +434,212 @@ impl BridgeTokenFactory {
...
+    Sets fee percentage for withdraw per silo per token
+    @param: fee_percent:- it should be with 6 decimals, eg: 10% = 0.1*10^6
+            silo_address:- should be an aurora account
+     */
+    #[access_control_any(roles(Role::FeeSetter))]
+    pub fn set_withdraw_fee_percentage_for_token_per_silo(
+        &mut self,
+        silo_address: &AccountId,
+        token: &String,
+        fee_percent: u128,
```
karim-en, line 572: Please use `U128` instead of `u128` for json input/output

bridge-token-factory/src/lib.rs
```diff
@@ -372,6 +670,41 @@ impl BridgeTokenFactory {
...
+                    eth_to_aurora: 0,
+                    eth_to_near: 0,
+                });
+
+        let amount_to_transfer: u128;
+        let mut fee_amount: u128;
+
+        // check if some msg attached than this call is for eth -> aurora otherwise eth -> near
+        match message.clone() {
+            Some(_message) => {
```
karim-en, line 686:
```suggestion
        match message {
            Some(_) => {
```
Extra clone

bridge-token-factory/src/lib.rs
```diff
@@ -78,6 +81,56 @@ pub enum Role {
...
     UnrestrictedDeployBridgeToken,
     MetadataManager,
+    FeeSetter,
+    FeeClaimer,
+}
+
+#[derive(
+    Default, BorshDeserialize, BorshSerialize, Debug, Clone, Serialize, Deserialize, PartialEq,
+)]
+pub struct DepositTokenBounds {
```
karim-en, line 91:
```suggestion
pub struct DepositFeeBounds {
```

bridge-token-factory/src/lib.rs
```diff
@@ -78,6 +81,56 @@ pub enum Role {
...
+    lower_bound: u128,
+    upper_bound: u128,
+}
+
+#[derive(
+    Default, BorshDeserialize, BorshSerialize, Debug, Clone, Serialize, Deserialize, PartialEq,
+)]
+pub struct WithdrawTokenBounds {
```
karim-en, line 99:
```suggestion
pub struct WithdrawFeeBounds {
```

bridge-token-factory/src/lib.rs
```diff
@@ -372,6 +670,41 @@ impl BridgeTokenFactory {
...
+        match deposit_fee_bound {
+            Some(token_bounds) => {
+                if fee_amount < token_bounds.lower_bound {
+                    fee_amount = token_bounds.lower_bound;
+                } else if fee_amount > token_bounds.upper_bound {
+                    fee_amount = token_bounds.upper_bound;
+                }
+            }
+            _ => {}
+        }
```
karim-en, line 705: This code for deposit and withdrawal could be moved to a separate function `adjust_value_between_bounds` that returns the adjusted fee value

bridge-token-factory/src/lib.rs
```diff
@@ -413,7 +753,62 @@ impl BridgeTokenFactory {
...
+            Some(token_bounds) => {
+                if fee_amount < token_bounds.lower_bound {
+                    //bound checks
+                    fee_amount = token_bounds.lower_bound;
+                } else if fee_amount > token_bounds.upper_bound {
+                    fee_amount = token_bounds.upper_bound;
+                }
+            }
+            _ => {}
+        }
```
karim-en, line 791: This code could be moved to the separate function `calculate_withdraw_fee`

karim-en: And the same for the deposit

bridge-token-factory/src/lib.rs
```diff
@@ -107,6 +160,16 @@ pub struct BridgeTokenFactory {
...
     #[deprecated]
     paused: Mask,
+    /// Fee amount bound for deposit.
+    pub deposit_fee_bound: UnorderedMap<String, DepositTokenBounds>,
+    /// Fee amount bound for withdraw
+    pub withdraw_fee_bound: UnorderedMap<String, WithdrawTokenBounds>,
+    /// Fee percentage of each token for deposit
+    pub deposit_fee_percentage: UnorderedMap<String, DepositFeePercentage>,
+    /// Fee percentage of each token for withdraw
+    pub withdraw_fee_percentage: UnorderedMap<String, WithdrawFeePercentage>,
```
karim-en, line 170: We could combine the `deposit_fee_bound` with `deposit_fee_percentage`, and `withdraw_fee_bound` with `withdraw_fee_percentage`. Because as I see, those values are used together in one call.

bridge-token-factory/src/lib.rs
UrAvgDeveloper: So we will only take fee as per silo, only for engine accounts which are configured with default or token-wise fee? If not set, we take fee whatever is set for aurora transfers.

bridge-token-factory/src/lib.rs
UrAvgDeveloper: fixed in 6a7df359a29f6105fae2cccc36ad5277fb747102

bridge-token-factory/src/lib.rs
UrAvgDeveloper: fixed in 6a7df359a29f6105fae2cccc36ad5277fb747102

bridge-token-factory/src/lib.rs
UrAvgDeveloper: fixed in 6a7df359a29f6105fae2cccc36ad5277fb747102

bridge-token-factory/src/lib.rs
UrAvgDeveloper: used single struct now, reducing code size. e53607f60125d25cb40abd0482f1120da1552d62

bridge-token-factory/src/lib.rs
UrAvgDeveloper: used single struct now, reducing code size. https://github.com/aurora-is-near/rainbow-token-connector/commit/e53607f60125d25cb40abd0482f1120da1552d62

bridge-token-factory/src/lib.rs
UrAvgDeveloper: fixed in e53607f60125d25cb40abd0482f1120da1552d62

bridge-token-factory/src/lib.rs
UrAvgDeveloper: made a generic function which uses enum. e53607f60125d25cb40abd0482f1120da1552d62

bridge-token-factory/src/lib.rs
UrAvgDeveloper: Can we include this in separate PR, as this will change the code a lot? Please let me know, I can start accordingly

bridge-token-factory/src/lib.rs
```diff
@@ -342,6 +427,242 @@ impl BridgeTokenFactory {
...
+        match self.withdraw_fee_percentage_per_silo.get(silo_address) {
+            Some(_silo_fee) => {
+                self.withdraw_fee_percentage_per_silo
+                    .get(silo_address)
+                    .unwrap()
+                    .token_to_fee_percentage
+                    .insert(token, &fee_percent);
+            }
+            None => {
+                let mut new_map: LookupMap<String, U128> = LookupMap::new(b"t");
```
karim-en, line 580: This logic is incorrect, this map has the same key, so it is shared between the tokens.
You can avoid using a sub map just by contacting `"token_id" + ":" + "silo_id"` and using it as a key.

bridge-token-factory/src/lib.rs
karim-en: yes. so I think that the `if withdraw_fee_percentage_per_silo.contain(withdrawer) {` is enough and better then `if is_aurora_engine_account(withdrawer.clone())`.
Actually, I think it can work as an exception list so that any account could be added to it, even some `DEX`

bridge-token-factory/src/lib.rs
karim-en: I think it is better to refactor it in this pr, so you just need to add an additional structs that will contain both `XxFeePercentage` and `FeeBounds `

bridge-token-factory/src/lib.rs
karim-en: It is not enough, all the fee calculations should be moved to separate methods, so the top-level function should contain just one line for the fee `let fee_amount = get_withdraw_fee(...)`

bridge-token-factory/src/lib.rs
UrAvgDeveloper: We need to have a `default_fee_percentage` for a silo as well, which is only used when there is no explicit fee set for a token wrt that silo. Should I just add this default fee in the map using silo address as key?

karim-en: Yes, you can use just the silo account id as a key

bridge-token-factory/src/lib.rs
karim-en: 382d2c2e632f5db6943c4753a9c7c8f397f411cd

bridge-token-factory/src/lib.rs
UrAvgDeveloper: 2e8a52d8469f1efb6fe551eb8eb3e5ee079df585

bridge-token-factory/src/lib.rs
karim-en: cce1ee19211a8a00b1a8567e867eb183c5e24e98

bridge-token-factory/src/lib.rs
karim-en: cce1ee19211a8a00b1a8567e867eb183c5e24e98

bridge-token-factory/src/lib.rs
```diff
@@ -342,6 +413,145 @@ impl BridgeTokenFactory {
...
+                self.withdraw_fee_percentage_per_silo
+                    .get(&get_silo_fee_map_key(&silo_account_id, None))
+            })
+    }
+
+    //this should be added as per: 10% -> 0.1 = 0.1*10^6
+    #[access_control_any(roles(Role::FeeSetter))]
+    pub fn set_deposit_fee(
+        &mut self,
+        token: &String,
```
karim-en, line 442: We need to convert the token address to lowercase, otherwise, it will be possible to pass the valid uppercase address

bridge-token-factory/src/lib.rs
```diff
@@ -342,6 +413,145 @@ impl BridgeTokenFactory {
...
+                },
+            },
+        );
+    }
+
+    //Fee should be added as per: 10% -> 0.1 = 0.1*10^6 with proper fee amount bounds
+    #[access_control_any(roles(Role::FeeSetter))]
+    pub fn set_withdraw_fee(
+        &mut self,
+        token: &String,
```
karim-en, line 468: We need to convert the token address to lowercase, otherwise, it will be possible to pass the valid uppercase address

bridge-token-factory/src/lib.rs
```diff
@@ -342,6 +413,145 @@ impl BridgeTokenFactory {
...
+    Sets fee percentage for withdraw per silo per token
+    @param: fee_percent:- it should be with 6 decimals, eg: 10% = 0.1*10^6
+            silo_account_id:- should be an aurora account
+            token:- token address, if passed `None` than default value is set.
+     */
+    #[access_control_any(roles(Role::FeeSetter))]
+    pub fn set_withdraw_fee_percentage_for_token_per_silo(
+        &mut self,
+        silo_account_id: AccountId,
+        token: Option<String>,
```
karim-en, line 500: We need to convert the token address to lowercase, otherwise, it will be possible to pass the valid uppercase address

bridge-token-factory/src/lib.rs
```diff
@@ -372,6 +582,10 @@ impl BridgeTokenFactory {
             target, message
         ));
 
+        let amount_to_transfer: u128;
+        let fee_amount = self.calculate_deposit_fee_amount(&token, amount, &message);
```
olga24912, line 586: fee amount calculated incorrectly: you should take percent from amount_to_transfer, not from amount.

bridge-token-factory/src/lib.rs
```diff
@@ -381,12 +595,18 @@ impl BridgeTokenFactory {
...
-            None => ext_bridge_token::ext(self.get_bridge_token_account_id(token))
+            None => ext_bridge_token::ext(self.get_bridge_token_account_id(token.clone()))
                 .with_static_gas(MINT_GAS)
                 .with_attached_deposit(env::attached_deposit() - required_deposit)
-                .mint(target, amount.into()),
+                .mint(target, amount_to_transfer.into())
+                .then(
+                    ext_bridge_token::ext(self.get_bridge_token_account_id(token))
+                        .with_static_gas(MINT_GAS)
+                        .with_attached_deposit(env::attached_deposit() - required_deposit)
```
olga24912, line 607: Not sure about that place... we have the same attached deposit to times in a raw. Don't we already spend it on the first call? Really not sure, how it works.

olga24912 (changes requested):

bridge-token-factory/src/lib.rs
karim-en: 13509deca41650d8e1c89c95f4dd10525e7a6fe4

bridge-token-factory/src/lib.rs
karim-en: 13509deca41650d8e1c89c95f4dd10525e7a6fe4

bridge-token-factory/src/lib.rs
```diff
@@ -342,6 +413,145 @@ impl BridgeTokenFactory {
...
+
+    pub fn get_withdraw_token_fee(&self, token: &String) -> Option<WithdrawFee> {
+        self.withdraw_fee.get(token)
+    }
+
+    // Returns fee percentage for different tokens per silo
+    pub fn get_withdraw_fee_percentage_per_silo(
+        &self,
+        silo_account_id: &AccountId,
+        token: &String,
```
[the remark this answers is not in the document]
karim-en: 13509deca41650d8e1c89c95f4dd10525e7a6fe4

bridge-token-factory/src/lib.rs
karim-en: feff748af2b5e00f67d773343188b219d4317590

bridge-token-factory/src/lib.rs
karim-en: The `amount_to_transfer` = `amount ` - `fee_amount `

bridge-token-factory/readme.md
```diff
@@ -0,0 +1,29 @@
...
+  * `set_deposit_fee_bound`: method to set upper and lower_bounds for deposit fee.
+  * `get_deposit_token_fee_percentage`: returns deposit token fee percentage for specific erc-20 token address passed.
+  * `get_deposit_token_fee_bound`: returns fee-bounds for deposits of specific erc-20 token address.
+
+
+* ### For withdraw:
+  
+  * `set_withdraw_fee_percentage`: method to set withdraw fee-percentage for near -> eth and aurora -> eth.
+  * `set_withdraw_fee_percentage_for_token_per_silo`: method to set withdraw fee-percentage for different aurora-silos per different tokens. Here first it checks for any default value set for the silo or not, than it updates the passed params fee-percentage values. 
+  * `set_default_withdraw_fee_percentage_per_silo`: method to set default fee-percentage for different aurora-silos per different tokens.
```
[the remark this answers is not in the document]
karim-en: 6df985a6d04df26648810784b830ac3f12f99fa2

bridge-token-factory/readme.md
```diff
@@ -0,0 +1,29 @@
...
+  * `set_deposit_fee_percentage`: method to set deposit fee-percentage for eth -> near and eth -> aurora.
+  * `set_deposit_fee_bound`: method to set upper and lower_bounds for deposit fee.
+  * `get_deposit_token_fee_percentage`: returns deposit token fee percentage for specific erc-20 token address passed.
+  * `get_deposit_token_fee_bound`: returns fee-bounds for deposits of specific erc-20 token address.
+
+
+* ### For withdraw:
+  
+  * `set_withdraw_fee_percentage`: method to set withdraw fee-percentage for near -> eth and aurora -> eth.
+  * `set_withdraw_fee_percentage_for_token_per_silo`: method to set withdraw fee-percentage for different aurora-silos per different tokens. Here first it checks for any default value set for the silo or not, than it updates the passed params fee-percentage values.
```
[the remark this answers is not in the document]
karim-en: 6df985a6d04df26648810784b830ac3f12f99fa2

bridge-token-factory/src/lib.rs
```diff
@@ -104,9 +154,12 @@ pub struct BridgeTokenFactory {
     pub owner_pk: PublicKey,
     /// Balance required to register a new account in the BridgeToken
     pub bridge_token_storage_deposit_required: Balance,
-    /// Mask determining all paused functions
-    #[deprecated]
-    paused: Mask,
+    /// deposit fee storage
```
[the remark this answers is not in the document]
karim-en: 13509deca41650d8e1c89c95f4dd10525e7a6fe4

bridge-token-factory/src/lib.rs
```diff
@@ -104,9 +154,12 @@ pub struct BridgeTokenFactory {
...
+    pub deposit_fee: UnorderedMap<String, DepositFee>,
+    /// withdraw fee storage
+    pub withdraw_fee: UnorderedMap<String, WithdrawFee>,
+    /// Withdraw fee percentage for each different silos for different tokens
```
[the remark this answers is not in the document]
karim-en: Refactored in 4f8905548ba0c948d8cc9c6fc6ce324bb1e266d0

bridge-token-factory/src/lib.rs
```diff
@@ -342,6 +413,145 @@ impl BridgeTokenFactory {
...
+        let Some(withdraw_fee) =
+            self.get_withdraw_token_fee(&token) else { return 0 };
+
+        // check whether the withdrawer is aurora native account or it's silos
+        let fee_amount = if let Some(silo_fee) =
+            self.get_withdraw_fee_percentage_per_silo(withdrawer, token)
+        {
+            (transfer_amount * silo_fee.0) / FEE_DECIMAL_PRECISION
+        } else {
+            (transfer_amount * withdraw_fee.fee_percentage.near_to_eth.0) / FEE_DECIMAL_PRECISION
```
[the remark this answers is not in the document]
karim-en: Refactored in 4f8905548ba0c948d8cc9c6fc6ce324bb1e266d0

bridge-token-factory/src/lib.rs
karim-en: It was also changed in 1b1d3fd1170eda4a4b7307c93bfda29533cd9f12.
So now for a deposit, there are no extra requirements for the storage deposit.
But it is not possible to attach a `deposit` for the withdrawal, so the storage deposit should be done while setting a fee for the token by `set_withdraw_fee`. 
Also, note that there is no storage deposit on calling `set_withdraw_fee_per_silo` because the token address could be none, so it is not possible to cover the storage deposit for all tokens when the fee is configured for all tokens in the silo. This case is fixed by adding a storage deposit in the `BridgeToken::new()` on deploying new tokens, but there is no solution for already deployed tokens, so this process should be done manually on migration.

bridge-token-factory/src/types.rs
```diff
@@ -0,0 +1,72 @@
...
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let s = <String as Deserialize>::deserialize(deserializer)?;
+        EthAddressHex::from_str(&s).map_err(|e| serde::de::Error::custom(e.0))
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
```
karim-en, line 22:
```suggestion
```

bridge-token-factory/readme.md
```diff
@@ -0,0 +1,28 @@
...
+1. To set fee-percentage and fee-bounds caller (predecessor) must have `FeeSetter` role.
+2. Fee-percentage has **6-decimal** precision for eg. to set deposit fee percentage as 10%, value for function param is 0.1 * 10^6 ie. 10^5.
+
+* ### For deposit: 
+  
+  * `set_deposit_fee_percentage`: method to set deposit fee-percentage for eth -> near and eth -> aurora.
+  * `set_deposit_fee_bound`: method to set upper and lower_bounds for deposit fee.
+  * `get_deposit_token_fee_percentage`: returns deposit token fee percentage for specific erc-20 token address passed.
+  * `get_deposit_token_fee_bound`: returns fee-bounds for deposits of specific erc-20 token address.
+
```
olga24912, line 14: Now we also have function `set_deposit_fee_per_silo` 

The README should be updated according current implementation.

bridge-token-factory/src/fee.rs
```diff
@@ -0,0 +1,661 @@
...
+                    upper_bound,
+                },
+            },
+        );
+
+        require!(env::attached_deposit() >= self.bridge_token_storage_deposit_required);
+        ext_bridge_token::ext(self.get_bridge_token_account_id(&token))
+            .with_static_gas(TOKEN_STORAGE_DEPOSIT_GAS)
+            .with_attached_deposit(env::attached_deposit())
+            .storage_deposit(None, None)
```
olga24912, line 47: What if we already have a registered token and we just want to update the data. We should prohibit update data or check if the token 

 Then either you need to explicitly prohibit updating the data, or check if the our account already registered for this token. 
 
 Do we refund excess deposit in this implementation?

bridge-token-factory/src/fee.rs
```diff
@@ -0,0 +1,661 @@
...
+    pub fn set_deposit_fee_per_silo(
+        &mut self,
+        silo_account_id: AccountId,
+        token: Option<EthAddressHex>,
+        fee_percentage: U128,
+        lower_bound: Option<U128>,
+        upper_bound: Option<U128>,
+    ) {
+        self.deposit_fee_per_silo.insert(
+            &get_silo_fee_map_key(&silo_account_id, token.as_ref()),
```
olga24912, line 61: Should we check that token exists in self.deposit_fee map first?

bridge-token-factory/src/fee.rs
```diff
@@ -0,0 +1,661 @@
...
+                    upper_bound,
+                },
+            },
+        );
+
+        require!(env::attached_deposit() >= self.bridge_token_storage_deposit_required);
+        ext_bridge_token::ext(self.get_bridge_token_account_id(&token))
+            .with_static_gas(TOKEN_STORAGE_DEPOSIT_GAS)
+            .with_attached_deposit(env::attached_deposit())
+            .storage_deposit(None, None)
```
olga24912, line 97: Same comment as for `set_deposit_fee` moreover, probable we already called `storage_deposit` for this token from `set_deposit_fee` function.

bridge-token-factory/src/fee.rs
```diff
@@ -0,0 +1,661 @@
...
+    ) -> Option<Fee> {
+        self.get_deposit_fee_per_silo_internal(&silo_account_id, token.as_ref())
+    }
+
+    // Accumulated fee should be claimed from here.
+    #[access_control_any(roles(Role::FeeClaimer))]
+    pub fn claim_fee(&self, token: AccountId, amount: Balance) {
+        ext_bridge_token::ext(token)
+            .with_static_gas(FT_TRANSFER_GAS)
+            .with_attached_deposit(1)
```
olga24912, line 153: Who provided this  One YoctoNEAR? Should this function be payable?

bridge-token-factory/src/lib.rs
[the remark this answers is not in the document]
olga24912: Anyway it will be good to add the encoding rules for keys in comments for each map.

bridge-token-factory/src/lib.rs
```diff
@@ -360,33 +400,57 @@ impl BridgeTokenFactory {
...
+        let Recipient { target, message } = parse_recipient(new_owner_id);
+
+        let token = EthAddressHex(token);
+        let amount_to_transfer: u128;
+        let fee_amount = self.calculate_deposit_fee_amount(
+            &token,
+            amount,
+            message.as_ref().and_then(|_| Some(&target)),
         );
+        amount_to_transfer = amount - fee_amount;
```
olga24912, line 412: Looks incorrect for me. 

Current fee calculation: 
fee_amount = amount *  fee_percentage =  (amount_to_transfer + fee_amount) *  fee_percentage

More logical: 
fee_amount = amount_to_transfer * fee_percentage
fee_amount + fee_amount*fee_percentage = amount * fee_percentage
fee_amount = amount * fee_percentage/(1 + fee_percentage)

bridge-token-factory/src/lib.rs
```diff
@@ -360,33 +400,57 @@ impl BridgeTokenFactory {
...
 
-        let Recipient { target, message } = parse_recipient(new_owner_id);
+        let multiplier = if fee_amount > 0 { 1 } else { 2 };
```
olga24912, line 414: If we transfer to aurora the multiplier should be 1 anyway, but we should add 1 yocto token

olga24912 (changes requested):

bridge-token-factory/src/lib.rs
olga24912: See comment here: https://github.com/aurora-is-near/rainbow-token-connector/pull/213#discussion_r1264986513

bridge-token-factory/src/fee.rs
karim-en: No, because this method can be used also to update the fee

bridge-token-factory/src/fee.rs
karim-en: Actually, there is a refund, but the problem is that the refund is done to the predecessor of the `storage_deposit` call, which is the factory contract.
I will try to figure out a solution for this.

bridge-token-factory/src/fee.rs
karim-en: 5c69b52e3017e6bcb8c0bf27f65f53065eed5158

bridge-token-factory/src/fee.rs
karim-en: Removed here bc39b1f074f2c9e0dfc2d3d0e76842ccd0b39e4b, because the storage deposit should be done manually while migration

bridge-token-factory/src/lib.rs
karim-en: Added 1 yocto, but the new storage deposit logic was reverted, so all storage deposit calls that related to the fee were removed. a6395e936c38f6adfcbb895c29c2b5cb5766b591.
The storage deposit is done for the factory account on deploying the new token. And storage deposits should be done manually for already exits tokens manually.

bridge-token-factory/readme.md
karim-en: 849520f130778a2f9f143d3b37d33d8299f2eda0

bridge-token-factory/src/lib.rs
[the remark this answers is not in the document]
olga24912: https://github.com/aurora-is-near/rainbow-token-connector/pull/213/commits/9a9e29de03ed5184f7a1f6fe39747a351c73f26e

bridge-token-factory/src/lib.rs
```diff
@@ -372,21 +414,61 @@ impl BridgeTokenFactory {
...
-                ),
-            None => ext_bridge_token::ext(self.get_bridge_token_account_id(token))
-                .with_static_gas(MINT_GAS)
-                .with_attached_deposit(env::attached_deposit() - required_deposit)
-                .mint(target, amount.into()),
+            Some(message) => {
+                let mint_self_promise =
+                    ext_bridge_token::ext(self.get_bridge_token_account_id(&token))
+                        .with_static_gas(MINT_GAS)
+                        .with_attached_deposit(self.bridge_token_storage_deposit_required)
```
olga24912, line 435: I don't like the logic with `storage_deposit`. Why the user should pay for the storage deposit for bridge factory? The Bridge Token registers the factory when it is created. We don't need to attach a deposit here.

bridge-token-factory/src/lib.rs
```diff
@@ -372,21 +414,61 @@ impl BridgeTokenFactory {
...
+                            .with_attached_deposit(self.bridge_token_storage_deposit_required)
+                            .mint(target, amount_to_transfer.into())
+                            .then(mint_fee_promise)
+                    } else {
+                        mint_fee_promise
+                    }
+                } else {
+                    ext_bridge_token::ext(self.get_bridge_token_account_id(&token))
+                        .with_static_gas(MINT_GAS)
+                        .with_attached_deposit(self.bridge_token_storage_deposit_required)
```
olga24912, line 468: If the target is already registered, we don't need to attach the deposit here. It will be good to add the possibility don't attach the near tokens for deposit.

olga24912 (changes requested):

bridge-token-factory/src/lib.rs
karim-en: This improvement is out of the scope of this PR, let's create an issue for that.
Also, note that in addition to this, we need to implement a refund mechanism.

bridge-token-factory/src/lib.rs
karim-en: In the current implementation, we already attach this amount of deposit, so we can keep it as it was for backward compatibility. Let's cover this improvement in a follow-up PR as described below.

olga24912 (approved):

PR #225, opened 2023-07-21
Silo to silo transfers

karim-en:
The PR implements a contract to make silo to silo transfers of the ERC20 tokens that are bridged from near.
The main public functions  to interact with the contract: 
```
registerToken(IEvmErc20 token) // To register the token address
storageDeposit(IEvmErc20 token, uint128 storageDepositAmount) // To do the storage deposit
ftTransferCallToNear(IEvmErc20 token, uint128 amount, string receiverId, string calldata message) // To initiate the silo to silo transfer
withdraw(IEvmErc20 token) // To do a refund if the transfer failed
```

To deploy the contract on testnet run: 
```
yarn hardhat deploy --silo aurora-testnet --network testnet_aurora
```
To upgrade the contract on testnet run: 
```
yarn hardhat upgrade --silo aurora-testnet --proxy 0x<PROXY_ADDRESS> --network testnet_aurora
```


silo-to-silo/contracts/SiloToSilo.sol
```diff
@@ -0,0 +1,182 @@
...
+    ) external {
+        string storage tokenAccountId = registeredTokens[token];
+        require(address(token) != address(0), "The token is not registered!");
+
+        token.transferFrom(msg.sender, address(this), amount);
+        // WARNING: The `withdrawToNear` method works asynchronously.
+        // As a result, there is no guarantee that this method will be completed before `initTransfer`.
+        // In case of such an error, the user will be able to call `withdraw` method and get his/her tokens back.
+        // We expect such an error not to happen as long as transactions were executed in one shard.
+        token.withdrawToNear(bytes(getNearAccountId()), amount);
```
olga24912, line 78: If we will fail here, than user balance will be = amount, but the tokens will be stuck on aurora contract and it will not be easy to withdraw.

silo-to-silo/contracts/SiloToSilo.sol
```diff
@@ -0,0 +1,182 @@
...
+        registeredTokens[token] = nearTokenAccountId;
+        emit TokenRegistered(token, nearTokenAccountId);
+    }
+
+    function ftTransferCallToNear(
+        IEvmErc20 token,
+        uint128 amount,
+        string calldata receiverId,
+        string calldata message
+    ) external {
```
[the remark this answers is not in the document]
karim-en: The external call usually is cheaper, more detail about the differences is here:
https://ethereum.stackexchange.com/questions/19380/external-vs-public-best-practices

The withdraw will be changed to external

silo-to-silo/contracts/SiloToSilo.sol
karim-en: It could fail just if the only if the account isn't registered, so that's why `registerToken` is exist

silo-to-silo/contracts/SiloToSilo.sol
```diff
@@ -0,0 +1,182 @@
...
+        PromiseCreateArgs memory callStorageDeposit = near.call(
+            nearTokenAccountId,
+            "storage_deposit",
+            args,
+            NEAR_STORAGE_DEPOSIT,
+            BASE_NEAR_GAS
+        );
+        callStorageDeposit.transact();
+
+        registeredTokens[token] = nearTokenAccountId;
```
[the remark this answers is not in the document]
karim-en: Double registration is not an issue in the trusted mode, even it better to have the possibility to call it again if something goes wrong, for example, passed a wrong address.
Anyway, I've added `// TODO: make it trustless` comment, so I think it is possible to make this call trustless.

silo-to-silo/contracts/SiloToSilo.sol
```diff
@@ -0,0 +1,182 @@
...
+    using AuroraSdk for PromiseCreateArgs;
+    using AuroraSdk for PromiseWithCallback;
+
+    bytes32 public constant CALLBACK_ROLE = keccak256("CALLBACK_ROLE");
+
+    uint64 constant BASE_NEAR_GAS = 10_000_000_000_000;
+    uint64 constant WITHDRAW_NEAR_GAS = 50_000_000_000_000;
+    uint64 constant FT_TRANSFER_CALL_NEAR_GAS = 200_000_000_000_000;
+
+    uint128 constant NEAR_STORAGE_DEPOSIT = 12_500_000_000_000_000_000_000;
```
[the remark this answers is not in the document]
karim-en: I had an idea to pass this value on `registerToken`, but this transfer is done just for the bridged tokens so they all have the same storage deposit.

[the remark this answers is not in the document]
karim-en: @olga24912 I forgot that for silo-to-silo transfer we can transfer not just the bridged token but any nep141 token, so I am going to add an additional argument

silo-to-silo/contracts/SiloToSilo.sol
```diff
@@ -0,0 +1,182 @@
...
+
+import {IERC20 as IERC20_NEAR} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {AuroraSdk, NEAR, PromiseCreateArgs, PromiseResultStatus, PromiseWithCallback} from "@auroraisnear/aurora-sdk/aurora-sdk/AuroraSdk.sol";
+import "@auroraisnear/aurora-sdk/aurora-sdk/Utils.sol";
+import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
+import "@openzeppelin/contracts/utils/Base64.sol";
+import "@openzeppelin/contracts/access/AccessControl.sol";
+import "./IEvmErc20.sol";
+
+contract SiloToSilo is AccessControl {
```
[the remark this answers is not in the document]
karim-en: 69796e73746aadc1b9620ca1385997f8def8d5ba

silo-to-silo/contracts/src/SiloToSilo.sol
```diff
@@ -0,0 +1,280 @@
...
+
+        string memory nearTokenAccountId = string(AuroraSdk.promiseResult(0).output);
+        registeredTokens[token] = TokenInfo(nearTokenAccountId, false);
+        emit TokenRegistered(token, nearTokenAccountId);
+    }
+
+    function storageDeposit(IEvmErc20 token, uint128 storageDepositAmount) external {
+        string storage tokenAccountId = registeredTokens[token].nearTokenAccountId;
+        require(bytes(tokenAccountId).length > 0, "The token is not registered!");
+
```
olga24912, line 87: If `registeredTokens[token].isStorageRegistered == true`, no need call `storage_deposit`. This check can safe some user wNEAR tokens.

silo-to-silo/contracts/src/SiloToSilo.sol
```diff
@@ -0,0 +1,280 @@
...
+    NEAR public near;
+    string public siloAccountId;
+
+    //[auroraErc20Token] => tokenAccountIdOnNear
+    mapping(IEvmErc20 => TokenInfo) public registeredTokens;
+
+    //[auroraErc20Token][userAddressOnAurora] => userBalance
+    mapping(IEvmErc20 => mapping(address => uint128)) public balance;
+
+    event TokenRegistered(IEvmErc20 token, string nearAccountId);
```
olga24912, line 44: Should we create more events? Also for storage deposit, withdraw and for ft_transfer_call?

silo-to-silo/contracts/src/SiloToSilo.sol
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    function withdrawCallback(address sender, IEvmErc20 token) external onlyRole(CALLBACK_ROLE) {
+        require(
+            AuroraSdk.promiseResult(0).status == PromiseResultStatus.Successful,
+            "ERROR: The `Withdraw` XCC is fail"
+        );
+
+        uint128 transferredAmount = _stringToUint(AuroraSdk.promiseResult(0).output);
+        balance[token][sender] -= transferredAmount;
```
olga24912, line 217: Let's decrease balance in `withdraw` and increase in `withdrawCallback` only if something goes wrong.

olga24912 (approved):

sept-en (changes requested):

silo-to-silo/contracts/src/SiloToSilo.sol
```diff
@@ -0,0 +1,280 @@
...
+        );
+
+        callFtTransfer.then(callback).transact();
+    }
+
+    function ftTransferCallCallback(address sender, IEvmErc20 token, uint128 amount) external onlyRole(CALLBACK_ROLE) {
+        uint128 transferredAmount = 0;
+        if (AuroraSdk.promiseResult(0).status == PromiseResultStatus.Successful) {
+            transferredAmount = _stringToUint(AuroraSdk.promiseResult(0).output);
+        }
```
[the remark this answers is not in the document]
karim-en: This case is handled here https://github.com/near/near-sdk-rs/blob/a33be41b38bc8ef7cc8fb7f81a1effc4803f2043/near-contract-standards/src/fungible_token/core_impl.rs#L186

silo-to-silo/contracts/src/SiloToSilo.sol
```diff
@@ -0,0 +1,348 @@
...
+
+    function _ftTransferCallToNear(
+        IEvmErc20 token,
+        string memory nearTokenAccountId,
+        uint128 amount,
+        string memory receiverId,
+        string memory message
+    ) private {
+        ftTransferCallCounter += 1;
+        emit InitFtTransferCall(ftTransferCallCounter, msg.sender, token, receiverId);
```
olga24912, line 200: Do we need to save also amount and message?

silo-to-silo/contracts/src/SiloToSilo.sol
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+    function isStorageRegistered(IEvmErc20 token) public view returns (bool) {
+        return registeredTokens[token].isStorageRegistered;
+    }
+
+    function getUserBalance(IEvmErc20 token, address userAddress) public view returns (uint128) {
+        return balance[token][userAddress];
+    }
+
```
olga24912, line 295: Maybe we also need a function `isRecipientStorageDeposited`?

silo-to-silo/contracts/src/SiloToSilo.sol
```diff
@@ -0,0 +1,348 @@
...
+    uint128 constant ASCII_9 = 57;
+    uint128 constant ONE_YOCTO = 1;
+    uint128 constant NO_DEPOSIT = 0;
+
+    NEAR public near;
+    string public siloAccountId;
+    uint256 public ftTransferCallCounter;
+
+    // auroraErc20Token => TokenInfo { nearTokenAccountId, isStorageRegistered }
+    mapping(IEvmErc20 => TokenInfo) public registeredTokens;
```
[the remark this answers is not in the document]
karim-en: We will add it in a separate PR

silo-to-silo/contracts/src/SiloToSilo.sol
```diff
@@ -0,0 +1,352 @@
...
+    event TokenRegistered(IEvmErc20 token, string nearAccountId);
+    event TokenStorageRegistered(IEvmErc20 token, string nearAccountId);
+    event RecipientStorageRegistered(IEvmErc20 token, string recipientId);
+    event Withdraw(IEvmErc20 token, address recipient, uint128 transferedAmount);
+    event InitFtTransferCall(uint256 nonce, address indexed sender, IEvmErc20 indexed token, string indexed receiverId);
+    event FtTransferCall(
+        uint256 nonce,
+        address indexed sender,
+        IEvmErc20 indexed token,
+        string indexed receiverId,
```
paouvrard, line 59:
```suggestion
    event InitFtTransferCall(uint256 indexed nonce, address indexed sender, IEvmErc20 indexed token, string receiverId);
    event FtTransferCall(
        uint256 indexed nonce,
        address indexed sender,
        IEvmErc20 indexed token,
        string receiverId,
```
Recommend indexing the nonce instead of receiverId, it will be useful to link the callback to the call. Also indexed strings don't work well for filtering: https://ethereum.stackexchange.com/questions/6840/indexed-event-with-string-not-getting-logged/7170#7170

silo-to-silo/contracts/src/SiloToSilo.sol
paouvrard: - I was going to say we can always get the amount from the Transfer event emitted in the same transaction but that wouldn't work with native ETH transfers (will this function be payable to add ETH support ?).
- The use case I had in mind for the `InitFtTransferCall` event is to be able to find the user's original transaction from `FtTransferCall` using the `nonce`. So imo it's ok not to have more parameters.

paouvrard (changes requested):

silo-to-silo/contracts/src/SiloToSilo.sol
karim-en: 82c51e23b2387c7329b8f81c57bd6a695c49caa0

silo-to-silo/contracts/src/SiloToSilo.sol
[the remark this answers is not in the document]
karim-en: f73b5035cec45edf1193f3e41e3e28e01fe3f91f

silo-to-silo/contracts/src/SiloToSilo.sol
```diff
@@ -0,0 +1,393 @@
...
+    ) public payable whenNotPaused {
+        require(near.wNEAR.balanceOf(address(this)) >= ONE_YOCTO, "Not enough wNEAR balance");
+        TokenInfo memory tokenInfo = registeredTokens[token];
+        require(tokenInfo.isStorageRegistered, "The token storage is not registered");
+
+        if (address(token) == address(0)) {
+            require(msg.value == amount, "Incorrect attached value");
+            _withdrawNativeTokenToNear(bytes(getImplicitNearAccountIdForSelf()), msg.value);
+        } else {
+            token.transferFrom(msg.sender, address(this), amount);
```
olga24912, line 192: Should we also check that `msg.value == 0` in that case?

olga24912 (changes requested):

silo-to-silo/contracts/src/SiloToSilo.sol
karim-en: d4d939d2dbc2c00fba2ced5a1ac656877c47a321

silo-to-silo/configs/aurora-testnet.json
```diff
@@ -0,0 +1,7 @@
+{
+    "siloAccountId": "aurora",
+    "nativeTokenAccountId": "aurora",
+    "auroraSdkAddress": "0x425cA8f218784ebE2df347E98c626094B63E7f30",
+    "auroraUtilsAddress": "0xc129336a6995F3b70A7139585403B82098260172",
```
olga24912, line 5: Why we don't have "auroraCodecAddress", but we have this field in aurora-mainnet.json

silo-to-silo/contracts/src/SiloToSilo.sol
```diff
@@ -0,0 +1,391 @@
...
+        uint128 storageDepositAmount,
+        string memory recipientId,
+        bool isRecipientStorageDeposit
+    ) private {
+        string memory nearTokenAccountId = registeredTokens[token].nearTokenAccountId;
+        require(bytes(nearTokenAccountId).length > 0, "The token is not registered");
+
+        PromiseCreateArgs memory callStorageDeposit = near.call(
+            nearTokenAccountId,
+            "storage_deposit",
```
olga24912, line 142: Do I understand correctly that for this to work you need to transfer tokens to the SiloToSiloImplicitNearAccountId? And we don't have any refund mechanism?

karim-en: cb8ff916385fd4675cb43bb4b95ded002e7bc44c

PR #226, opened 2023-08-01
Silo to silo tests

olga24912:
* Bugs fixes in `SiloToSilo.sol`:
    * Decrease gas for ft_transfer_call for not exceeding the gas limit 
    * Fix check that the token is not registered
    * Fix arguments for ftTransferCallToNearCallback
    * Fix arguments for withdraw call
    * Remove redundant check of balance
    * Fix the string to Uint function. The `AuroraSdk.promiseResult(0).output` can return `"100"` not just `100`. 
* Before running tests don't forget about git submodule. I add this https://github.com/aurora-is-near/aurora-contracts-sdk.git repo as a submodule for using some Sol libs such as `AuroraSdk.sol`. Let me know if you can do it a better way. 
*  Tests in file `integration-tests/src/lib.rs`: 
    *  `test_ft_transfer_to_silo` -- transfer mock_tokens from engine to silo and back
    * `test_withdraw` -- the transfer is failed because of silo account is not registered and withdraw method executed 
    * `check_access_control` -- checks that account without access can't run token registration and callback functions
    * `error_on_withdraw_to_near` -- create a mock token where for user registration we need to deposit 10 NEAR tokens, register tokens in our contract, but really it is not registered, try to transfer these tokens and it fails during operation `token.withdrawToNear(bytes(getNearAccountId()), amount);` . Try to withdraw user tokens back. This test fails due to an issue in our contract.


silo-to-silo/integration-tests/aurora-sdk-integration-tests/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "aurora-sdk-integration-tests"
```
karim-en, line 2: Let's contribute to the sdk repo instead of copy-pasting it

silo-to-silo/contracts/src/SiloToSilo.sol
```diff
@@ -174,9 +172,13 @@ contract SiloToSilo is AccessControl {
 
     function _stringToUint(bytes memory b) private pure returns (uint128) {
         uint128 result = 0;
+        
         for (uint256 i = 0; i < b.length; i++) {
-            result = result * 10 + (uint128(uint8(b[i])) - 48);
+            if (uint8(b[i]) > uint8(47) && uint8(b[i]) < uint8(58)) {
```
karim-en, line 177: Let's assign the `b[i]` to a variable
`uint256 v = uint256(uint8(b[i]));`

also `47` and `58` confuse me so I think it is better to use >= 48 && <=57

silo-to-silo/contracts/src/SiloToSilo.sol
```diff
@@ -174,9 +172,13 @@ contract SiloToSilo is AccessControl {
 
     function _stringToUint(bytes memory b) private pure returns (uint128) {
```
karim-en, line 173: I propose adding this function to the SDK, I think this will useful

karim-en (changes requested):

silo-to-silo/integration-tests/aurora-sdk-integration-tests/Cargo.toml
olga24912: fixed: https://github.com/aurora-is-near/rainbow-token-connector/pull/226/commits/585b9859142a93201ec16a5f49c75a80f8159be6

silo-to-silo/contracts/src/SiloToSilo.sol
olga24912: fixed: https://github.com/aurora-is-near/rainbow-token-connector/pull/226/commits/82d74e3178cc9e3bb9e292cf6b286722d6338591

silo-to-silo/contracts/src/SiloToSilo.sol
olga24912: I'm not sure this is the right place. This is parsing of a specific output result.....

Anyway, I think it should be in the separate PR.

karim-en (approved):

PR #232, opened 2023-10-09
Bridge factory & tokens: add  `attach_full_access_key` api

karim-en:
This PR adds the possibility of adding a full access key by DAO as a fallback.
- Added `pub attach_full_access_key` to the factory contract and limited its access by role `Role::DAO`
- Added `pub attach_full_access_key` to the bridge token contract and limited its access by `controller_or_self`
- Added `pub attach_full_access_key_to_token` to the factory contract and limited its access by role `Role::DAO`, this method call the `attach_full_access_key` for the bridge token.
- Updated the `near-workspace` crate to be able to use `view_access_keys`.
- Updated the rust toolchain due to the new  `near-workspace` crate.
- Bump contracts versions to `0.2.1`


vzctl (approved):

bridge-token-factory/tests/token_transfer.rs
```diff
@@ -669,3 +651,96 @@ async fn test_upgrade() {
...
+    let keys: Vec<String> = token_account
+        .view_access_keys()
+        .await
+        .unwrap()
+        .iter()
+        .map(|info| info.public_key.to_string())
+        .collect();
+
+    // Check that the access key is added to the bridge token contract
+    assert!(keys.len() == 1 && keys.contains(&ed_pk_str.to_string()));
```
olga24912, line 745: Optional, may be it can be a good idea also add test, which checks that account without access right cannot add full access key.

olga24912 (approved):

PR #245, opened 2024-05-30
Minor audit fixes

karim-en:
* Remove unused import `Ownable.sol`.
* Remove unused `_pausedFlags`.
* Change withdraw argument type from u256 to u128 to avoid potential overflow.
* Replace `OwnableUpgradeable` with `Ownable2StepUpgradeable`


kiseln (approved):
All minor issues but reasonable

PR #248, opened 2024-08-22
Upgrade the bridge to work with chain signatures

kiseln:
Done:
- ProofConsumer deleted
- Factory is now initialized with a derived EVM address of a Near bridge contract
- Deploy token changed to verify signature instead of Near proof
- Deposit changed to verify signature instead of Near proof
- Nonce validation performed to prevent replay attacks

Future work:
- Fees?
- Add Wormhole version for withdrawals


erc20-bridge-token/contracts/BridgeTokenFactory.sol
```diff
@@ -35,12 +35,30 @@ contract BridgeTokenFactory is
     bool private _isWhitelistModeEnabled;
 
     address public tokenImplementationAddress;
+    address public nearBridgeDerivedAddress;
+
+    mapping(uint128 => bool) private _completedTransfers;
```
karim-en, line 40: Let's make it public so the front can check if it is completed or not.

karim-en (changes requested):

PR #253, opened 2025-02-04
Improve the Omni api integration

karim-en:


olga24912 (approved):

karim-en:
> What is `deploy_token` for? What tokens will use `finish_withdraw` and `finish_withdraw2`?
- `deploy_token` to keep the `factory.bridge.near` as a deployer for Ethereum tokens
- `finish_withdraw` will be used after upgrading the factory but before upgrading the tokens.
- `finish_withdraw_2` will be used after upgrading the tokens.

